PR #9, opened 2019-08-21
feat: add cli support for launch in commander.

brizer:
Added functionality for use in the terminal. For example, I can use it in my bash, simple use `npmgraph request`, then it will start a server build on your static files, and open browser automatic. It becomes very convenient and fast. I also intend to add the dependency analysis of local project which is unpublished.

### DIFF
--- a/bin/npmgraph.js
+++ b/bin/npmgraph.js
@@ -5,19 +5,29 @@ const open = require('open');
 const argv = require('minimist')(process.argv.slice(2));
 
 const app = express();
-const port = 3003;
+const port = process.env.PORT || argv.port || 3003;
 let packageName = '';
 
 const getPackageName = (packlist = []) => {
   if (packlist.length > 0) return packlist[0];
-  console.log('Please assign packageName');
 };
 
+function usage() {
+  console.log(`Usage: npmgraph [--port=portnum] package_name
+  --port=portnum:  server port to listen on';
+  package_name: package to start dependency graph at`);
+}
+
 //handle cli params
-if (argv.local) {
-} else {
-  packageName = getPackageName(argv._);
-  if (!packageName) return;
+packageName = getPackageName(argv._);
+if (!packageName) {
+  usage();
+  process.exit(1);
+}
+//use remote mode
+if (argv.r) {
+  open(`http://npm.broofa.com?q=${packageName}`);
+  process.exit(1);
 }
 
 //view engine setup
@@ -29,5 +39,5 @@ app.use(express.static(path.join(__dirname, '../')));
 
 app.listen(port, () => {
   console.log(`npmgraph listening on port ${port}!`);
-  open('http://localhost:3003');
+  open(`http://localhost:${port}`);
 });
--- a/bin/npmgraph.js
+++ b/bin/npmgraph.js
@@ -13,11 +13,11 @@ const getPackageName = (packlist = []) => {
 };
 
 function usage() {
-  console.log(`Usage: npmgraph [--port=portnum] package_name
+  console.log(`Usage: npmgraph [--port=portnum] package_name [-r]
   --port=portnum:  server port to listen on';
-  package_name: package to start dependency graph at`);
+  package_name: package to start dependency graph at;
+  -r: use remote mode to show graph`);
 }
-
 //handle cli params
 packageName = getPackageName(argv._);
 if (!packageName) {
--- a/bin/npmgraph.js
+++ b/bin/npmgraph.js
@@ -1,0 +1,33 @@
+#!/usr/bin/env node
+const path = require('path');
+const express = require('express');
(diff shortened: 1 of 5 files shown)


bin/npmgraph.js
```diff
@@ -0,0 +1,33 @@
...
+const port = 3003;
+let packageName = '';
+
+const getPackageName = (packlist = []) => {
+  if (packlist.length > 0) return packlist[0];
+  console.log('Please assign packageName');
+};
+
+//handle cli params
+if (argv.local) {
```
broofa: remove?  (arv.local not referenced elsewhere)

bin/npmgraph.js
```diff
@@ -0,0 +1,33 @@
...
+} else {
+  packageName = getPackageName(argv._);
+  if (!packageName) return;
```
broofa: Bash scripts should exit with a non-zero code anytime there's an error.  Also, error and action related to this should happen together.

```suggestion
 if (!packageName) {
   console.log('Please assign packageName');  // or usage()? see below
   process.exit(1);
 }
```

... and actually even that can be improved. If we're adding support for a `--port` flag we should document that it's a thing, so have a usage function like so:
```
function usage() {
  console.log(`Usage: npmgraph [--port=portnum] package_name
  --port=portnum:  server port to listen on');
  packaage_name: package to start dependency graph at`);
}
```

bin/npmgraph.js
```diff
@@ -0,0 +1,33 @@
...
+const open = require('open');
+const argv = require('minimist')(process.argv.slice(2));
+
+const app = express();
+const port = 3003;
+let packageName = '';
+
+const getPackageName = (packlist = []) => {
+  if (packlist.length > 0) return packlist[0];
+  console.log('Please assign packageName');
```
broofa: It's weird that this is logged here, but that the actual action related to this (exiting the program) happens elsewhere.  See comment below.

broofa:
All of the above comments aside, why not just have this be a shell function that opens the page on npm.broofa.com? ...
```
function npmgraph() {
  open "http://npm.broofa.com?q=$*"
}
```

It avoids the hassle (for the user) of having to think about the fact there's a server running locally, how and when to kill it, and causing page reloads to stop working when the server goes away, etc.

bin/npmgraph.js
brizer: Thanks, I have learned a lot.

brizer:
https://github.com/broofa/npmgraph/pull/9#issuecomment-523459629, this is a good idea. But if I want to analysis a local project which is not published, what should I do?

brizer:
For https://github.com/broofa/npmgraph/pull/9#issuecomment-523459629, I think provide a param -r means remote, it's better. Do you agree?

broofa:
> But if I want to analysis a local project which is not published, what should I do?

Can you elaborate on how running a local server helps with this?  Does it resolve `file:` dependencies properly?  I have to confess I've never tested this.

brizer:
> > But if I want to analysis a local project which is not published, what should I do?
> 
> Can you elaborate on how running a local server helps with this? Does it resolve `file:` dependencies properly? I have to confess I've never tested this.
> 
> [Edit: Oh, I see you said you intend to add that... cool!]


Now, we can use `--port=** ` to choose port number and `-r` to choose remote model. Then, we can add a param such as `-l` means local project, to analysis local project. But we have to change the source of npmgraph. I think it's more readable than `npm ls`

brizer:
> [Just a heads up that I'm going on vacation so I may be slow in responding here over the next week or so.]
> 
> My apologies for the longish comment here. 'Doing a bit of brainstorming on this. Feel free to push back...
> 
> It's unfortunate that this command blocks the terminal. It'd be nice if it spun up the server in the background and returned immediately so you could use it just like any (most) other CLIs. I think that's doable with the following changes...
> 
> 1. Launch the server as a separate process (using `spawn` + [`detached: true`](https://nodejs.org/api/child_process.html#child_process_options_detached) ?)
> 2. Have the spawned server gracefully handle the case where a server may already be running. (`app.listen()` throws EADDRINUSE).  It should just quietly exit.
> 3. Provide some mechanism for tearing down the server
> 
> The third point is actually rather problematic. Either the server quits on it's own (leaving the user with a webpage that breaks if refreshed) or it stays up (coming cpu & memory resources that are unlikely to be needed.)
> 
> .... and I gotta pack and head out. Sorry to leave this dangling, but I wanted to share this thought.
> 
> Thanks again for helping out!




Thanks for sharing your thought. But I have a question, which cli tool can spun up the server in the background and returned immediately? As far as I can see, a node server process which based on express such as webpack-dev-server, also blocks the terminal. Could you show me some example? Thanks again.

brizer:
EEE..... What is going on? My friend?

PR #17, opened 2020-02-05
Feature: add file picker as an upload option

garyg1:
### Changes
- add file-picker alongside drag-to-upload box

### Comments
I usually fullscreen Chrome on Mac, so the lack of a file picker option to upload my `package.json` caused bad UX for me. Currently, in order to upload my package.json, I had to
1. un-fullscreen chrome,
2. find the file using Finder,
3. drag it into the application
4. and then re-fullscreen chrome.

I like the idea of the drag box, but I think it would be better UX to also give users the option to use the browser's picker.

### Testing
I hand-tested that the drag-box and the new upload prompt worked on the latest versions of Chrome (Mac), Safari (Mac), Firefox (Ubuntu 18.04), and Edge (Windows 10).

#### New Upload View
Screen Shot 2020-02-05 at 12 33 02 PM

#### Current Upload View (for comparison)
Screen Shot 2020-02-05 at 12 33 09 PM

### DIFF
--- a/index.css
+++ b/index.css
@@ -231,6 +231,7 @@ g .stub > path {
   background-color: #ddd;
   border: dashed 2px #888;
   padding: 1em;
+  cursor: pointer;
 }
 #drop_target.drag {
 	background-color: #ccf;
--- a/index.html
+++ b/index.html
@@ -91,23 +91,9 @@
         <p>
         Enter NPM module name here <i class="material-icons">arrow_upward</i> to see the dependency graph.  To graph multiple modules separate names with commas (e.g. <a href="?q=mocha, chalk, rimraf">"mocha, chalk, rimraf"</a>).
         </p>
-        <br/>
         <div id="drop_target" style="text-align: center">
-					... drop a <code>package.json</code> file here
+          ... or drop a <code>package.json</code> file here
         </div>
-        <br/>
-        <p>
-          <div style="text-align: center;">
-            ... or upload a <code>package.json</code> file here
-            <p>
-              <form id="upload_file" style="text-align: center;">
-                <input type="file" name="file" />
-                <input type="submit" value="Done" />
-              </form>
-            </p>
-          </div>
-          
-        </p>
         <div id="info-footer">
           <div id="copyright">
             &copy; Robert Kieffer, 2017  MIT License
--- a/index.html
+++ b/index.html
@@ -91,9 +91,23 @@
         <p>
         Enter NPM module name here <i class="material-icons">arrow_upward</i> to see the dependency graph.  To graph multiple modules separate names with commas (e.g. <a href="?q=mocha, chalk, rimraf">"mocha, chalk, rimraf"</a>).
         </p>
+        <br/>
         <div id="drop_target" style="text-align: center">
-					... or drop a <code>package.json</code> file here
-				</div>
+					... drop a <code>package.json</code> file here
+        </div>
+        <br/>
+        <p>
+          <div style="text-align: center;">
+            ... or upload a <code>package.json</code> file here
+            <p>
+              <form id="upload_file" style="text-align: center;">
+                <input type="file" name="file" />
+                <input type="submit" value="Done" />
+              </form>
+            </p>
+          </div>
+          
+        </p>
         <div id="info-footer">
           <div id="copyright">
             &copy; Robert Kieffer, 2017  MIT License
--- a/index.html
+++ b/index.html
@@ -94,6 +94,7 @@
         <div id="drop_target" style="text-align: center">
           ... or drop a <code>package.json</code> file here
         </div>
+        <input id="file_upload" type="file" accept="application/json" style="position: fixed; top: -100em" />
         <div id="info-footer">
           <div id="copyright">
             &copy; Robert Kieffer, 2017  MIT License
--- a/js/index.js
+++ b/js/index.js
@@ -65,6 +65,23 @@ function zoom(op) {
   }
 }
 
+/** Loads a `File` object as a package.json. */
+async function graphPackageJSON(file) {
+  if (!file) return alert('Please drop a file, not... well... whatever else it was you dropped');
+  if (file.type && file.type != 'application/json') return alert('File must have a ".json" extension');
+
+  const reader = new FileReader();
+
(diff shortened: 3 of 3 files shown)


js/index.js
```diff
@@ -239,6 +239,38 @@ onload = function() {
   Inspector.init();
   Inspector.showPane('pane-info');
 
+  /** Loads a `File` object as a package.json. */
+  async function loadPackageJSON(file) {
```
broofa: I believe this function can be pulled out to top level of this file. (It doesn't reference any graph()-scoped vars here, does it?)  If so, just move to the end of this file.

(And, yes, I know this codebase is a bit of a mess overall so I probably shouldn't get too picky.  This whole thing was a weekend-project that I've never really gotten around to cleaning up.  But... what's a guy to do?  :-) )

js/index.js
```diff
@@ -239,6 +239,38 @@ onload = function() {
...
+      ev.preventDefault();
+
+      // get file from upload `input`
+      const formData = new FormData(uploadFileRef);
+      const file = formData.get('file');
+      if (!file) return alert('No files found in file upload');
+
+      return loadPackageJSON(file);
+    }
+  });
```
broofa: Pairing with changes above, just watch for change event on the file input?  (Haven't actually tested this, but I think this should work, no?)
```
$('#upload_file').onchange = ev => {
  ev.preventDefault();
  (... etc.)
}
```

index.css
```diff
@@ -231,6 +231,7 @@ g .stub > path {
   background-color: #ddd;
   border: dashed 2px #888;
   padding: 1em;
+  cursor: pointer;
```
garyg1: Hint to the user that they can also click to upload.

js/index.js
```diff
@@ -239,8 +256,27 @@ onload = function() {
...
   Inspector.showPane('pane-info');
 
+
+  // Handle file upload
+  $('#file_upload').onchange = async changeEv => {
+    changeEv.preventDefault();
+
+    const files = changeEv.target.files;
+    if (!files || files.length !== 1) {
+      return;
```
garyg1: No error message, because it might be the case that there is currently 1 file, and the user clicks again and cancels.

index.html
```diff
@@ -92,8 +92,9 @@ <h2>package.json<span class="material-icons"></span></h2>
         Enter NPM module name here <i class="material-icons">arrow_upward</i> to see the dependency graph.  To graph multiple modules separate names with commas (e.g. <a href="?q=mocha, chalk, rimraf">"mocha, chalk, rimraf"</a>).
         </p>
         <div id="drop_target" style="text-align: center">
-					... or drop a <code>package.json</code> file here
-				</div>
+          ... or drop a <code>package.json</code> file here
+        </div>
+        <input id="file_upload" type="file" accept="application/json" style="position: fixed; top: -100em" />
```
garyg1: The `-100em` is from [here](https://stackoverflow.com/a/28075416/8887313) on Stack Overflow. Apparently, using `display: none` may cause the input to be disabled in some browsers. (Best not to have things potentially break in the future.)

index.html
```diff
@@ -91,9 +91,23 @@ <h2>package.json<span class="material-icons"></span></h2>
...
+            ... or upload a <code>package.json</code> file here
+            <p>
+              <form id="upload_file" style="text-align: center;">
+                <input type="file" name="file" />
+                <input type="submit" value="Done" />
+              </form>
+            </p>
+          </div>
+          
+        </p>
```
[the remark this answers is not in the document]
garyg1: Thanks -- addressed this in the changes.

index.html
```diff
@@ -91,9 +91,23 @@ <h2>package.json<span class="material-icons"></span></h2>
         <p>
         Enter NPM module name here <i class="material-icons">arrow_upward</i> to see the dependency graph.  To graph multiple modules separate names with commas (e.g. <a href="?q=mocha, chalk, rimraf">"mocha, chalk, rimraf"</a>).
         </p>
+        <br/>
```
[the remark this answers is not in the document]
garyg1: Fixed!

js/index.js
garyg1: Agreed -- fixed.

js/index.js
garyg1: Implemented this in the new changes.

js/index.js
broofa: This is the sort of thing that's good to have as a comment in the code.

index.html
broofa: Also, comment in code (w/ link to SO article)

broofa:
What's the reason for hiding the input button?  Hiding it seems a little too... well... hidden.  And leads to hacks like the -100 margin, which invariably have edge case issues (e.g. elements that aren't visible, but that can be tab-navigated to are always a little weird.)

garyg1:
Thanks for the feedback -- I agree, I didn't take tab-navigation or accessibility into account. Also sorry, I should have said this up before changing everything.

I wanted to hide the `<input>` with no "confirm" button because of these reasons. (I realize that the hidden `<input>` has some of these same problems, anyway.)

- it looks weird with the centered layout
Screen Shot 2020-02-17 at 10 36 39 AM

- I feel like it's best not do things with their file without user consent, and for an `<input type="file">` I don't think it's obvious that something will happen with their file after `onchange`.

So I think the options would be to
1. Use a visible `<input>` with no button and listen to `onchange`
2. Use a visible `<input>` with a confirm button and listen for the button `onclick`
3. Move the hidden `<input>` to a better place, which would ensure it doesn't interfere with accessibility features like tab-navigation.

Thanks for the help with this PR -- what would be your recommendation on how to reimplement this correctly?

fregante:
I haven't reviewed the code, but I have couple of notes:

- The native `input[file]` widget already supports drag and drop, but I'm not sure of how cross-compatible it is
- You can merge the 2 ways into a single visual area like GitHub does in comments and releases: 
	Screen Shot
- The whole page can be a dropzone, like Google Drive and Dropbox for example

fregante:
I reimplemented in #55 to avoid conflicts and use a single drop/select area.

garyg1:
Closing in favor of #55.

@broofa I want to apologize for my unproductive discussion. Thank you for your time reviewing this.

PR #24, opened 2020-09-29
show unpackedSize in module detail

chuongtrh:


js/Inspector.js
```diff
@@ -240,6 +240,7 @@ export default class Inspector {
       const popularity = scores ? (scores.popularity * 100).toFixed(0) + '%' : 'n/a';
       const maintenance = scores ? (scores.maintenance * 100).toFixed(0) + '%' : 'n/a';
       let license = module.licenseString;
+      const unpackedSize = pkg.dist.unpackedSize ? formatBytes(parseInt(pkg.dist.unpackedSize)) : 0;
```
styfle, line 243:
```suggestion
      const unpackedSize = pkg.dist.unpackedSize ? formatBytes(parseInt(pkg.dist.unpackedSize)) : '';
```

styfle (approved):

broofa:
This looks interesting, thanks for the contribution!

Quick question... what is `unpackedSize`?  And is meaningful in a way developers are likely to care about?

For example, https://registry.npmjs.org/uuid/8.3.0 claims an `unpackedSize` of 112959 bytes.  But if install the package and do `du -sh node_modules/uuid`, I shows it's actually 304K on disk.  But nearly all of that is examples or maybe test code.   If you bundle it in a project using webpack it typically only adds 1-2KB to the bundle.  So... what does that 113K even mean?

chuongtrh:
Thank you for your response!

**unpackSize** is a measurable metric to see how large a package.
My idea inspired from this blog [https://medium.com/wix-engineering/keep-your-bundle-size-under-control-with-import-cost-vscode-extension-5d476b3c5a76](https://medium.com/wix-engineering/keep-your-bundle-size-under-control-with-import-cost-vscode-extension-5d476b3c5a76) (Author extension **Import Cost** on VSCode)

broofa:
> unpackSize is a measurable metric to see how large a package.

What is it measuring, though?  That blog post you link to describes a tool that shows a value with a well-articulated meaning.  It shows the amount of code that is bundled given a specific import statement.  That has tangible value to developers.  I understand that.  They understand that.

But `unpackedSize` doesn't provide that.  It's measuring something else... and I don't know what that is.  For example, take a gander info at https://registry.npmjs.org/uuid/latest. Where is the unpackedSize value of "112959" coming from?  It's not the size of the compressed tarball (27105).  It's not the size of the uncompressed tarball (165888).  Nor is it the size of the extracted files (304K).  And it's certainly not the size of [the `import`ed size of code](https://github.com/uuidjs/uuid/blob/master/bundlewatch.config.json).

I'm not comfortable calling out this value in the UI when I don't know how to explain what it is because I'm going to get questions about it.

chuongtrh:
I am no idea how to explain. I try to find an answer on the npmjs.com document but can not find it.

> The "Unpacked Size" (basically Publish Size) is available on the npmjs.com website
[https://stackoverflow.com/a/52207709/5589964](https://stackoverflow.com/a/52207709/5589964)

styfle:
I would describe npmjs.com "Unpacked Size" as "Publish Size", the files published to npm before being compressed to a tarball (or perhaps after being extracted from the tarball).






The value is nearly identical to the "Publish Size" reported by [Package Phobia](https://packagephobia.com).

However, I don't know if npm's algorithm is open source, but Package Phobia is [here](https://github.com/styfle/packagephobia).

Example: https://packagephobia.com/result?p=uuid
API: https://packagephobia.com/v2/api.json?p=uuid

broofa:
The fact we already have disagreement over the definition is, imho, a pretty good indicator of the issue this poses.

> I see unpackedSize a metric to see **the cost of adding a new dependency to a project**

... versus ...

> I would describe npmjs.com "Unpacked Size" as "Publish Size", **the files published to npm before being compressed** to a tarball (or perhaps after being extracted from the tarball).

Comparing some packages off the top of my head (not cherry-picking, these are the first three that came to mind) and compare unpackedSize to bundle size and installed files size, I find the following: 

| package | unpackedSize | bundle size (bundlephobia.com) | size of installed files |
|---|---|---|---|
|axios| 366K |13K (self) + no dependencies|476K|
|request| 209K | 44K (self) + 640K in dependencies | 328K |
|express| 208K | 34K (self) + 507K in dependencies | 476K |

Based on this, here are my concerns with `unpackedSize`:
1. It doesn't correlate well to any intuitive definition of "package size"
1. It is particularly bad at predicting bundle size, which is what most developers are going to care about
1. Actual cost of a package depends on how many dependencies it shares with other packages in a project.  E.g. The cost of installing `request` falls if you already have `express` installed and it can share the `mime-db` dependency.

Given the above, I'm reticent to make this property a prominent part of the UI.  It's simply not informative enough to warrant that.  If you want, we can whitelist the "dist" property [here](https://github.com/broofa/npmgraph/blob/master/js/Inspector.js#L214-L217) so it's displayed in the "package.json" section of the inspector when you click on a module in the graph.

I would love to add a feature that allows developers to see the "cost" of packages, but it's a non-trivial problem.   To do that properly you need to display the _incremental_ cost of adding a package.  I.e. Display the bundle-size of the package plus the bundle size of any _new_ (exclusive) dependencies it adds.  E.g. Adding `request` to an `express` project reflect the fact `request` can share `express`' dependency on `mime-types`.

broofa:
Note: It looks like bundlephobia has an API for getting bundle size and composition. E.g. https://bundlephobia.com/api/size?package=express@4.17.1

This could enable the incremental cost feature I just mentioned.  E.g. When you select a module in the graph, the "Module" tab in the inspector could show three sizes:
* "self" (size of code directly included in module)
* "shared dependencies" (total size of dependencies module shares with other modules in the graph)
* "exclusive dependencies" (total size of dependencies only reachable through target module - i.e. the *actual* cost of adding the module to the project.)

chuongtrh:
Three metric above look like more clear than unpackedSize. Hope to see it in "Module" tab

Thanks @styfle @broofa

broofa:
Created a new issue for the bundlephobia feature I described.

PR #41, opened 2021-01-05
Avoid shrinking tabs on large dependency trees

fregante:
gif

1. Visit http://npm.broofa.com/?q=webpack
2. Resize the window vertically
3. Notice the tabs getting squished

### DIFF
--- a/index.css
+++ b/index.css
@@ -193,6 +193,7 @@ summary {
 
 #tabs {
   flex-basis: auto;
+  flex-shrink: 0;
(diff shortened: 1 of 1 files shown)


broofa:
Good catch, thanks!  (I'm not in the habit of testing Safari or FF, clearly.)

BTW, how are you creating these animated screen captures?  They're really helpful.

fregante:
https://www.cockos.com/licecap/

But there's also https://getkap.co for higher quality captures

PR #49, opened 2021-02-04
Fix module heading link

fregante:
`href="#"` never makes sense; It should either use a real `href` or be a `button`

Untested

### DIFF
--- a/js/Inspector.js
+++ b/js/Inspector.js
@@ -23,10 +23,11 @@ export function ExternalLink({ href, children, target = '_blank', className, sty
 export function QueryLink({ query }) {
   const { query: [, setQuery] } = useContext(AppContext);
   if (!Array.isArray(query)) query = [query];
-  return html`<a href="#" onClick=${e => {
+  const url = location.pathname}?q=${query.join(',');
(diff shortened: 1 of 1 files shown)


js/Inspector.js
```diff
@@ -23,10 +23,11 @@ export function ExternalLink({ href, children, target = '_blank', className, sty
 export function QueryLink({ query }) {
   const { query: [, setQuery] } = useContext(AppContext);
   if (!Array.isArray(query)) query = [query];
-  return html`<a href="#" onClick=${e => {
+  const url = location.pathname}?q=${query.join(',');
+  return html`<a href=${url} onClick=${e => {
```
fregante, line 27: Also I suggest this wrapping the whole listener with: https://github.com/fregante/filter-altered-clicks

```suggestion
  return html`<a href=${url} onClick=${filterAlteredClicks(e => {
```

Or this if you don't want an extra dependency:
```suggestion
  return html`<a href=${url} onClick=${e => {
    if (
      e.which > 1 ||
      e.shiftKey ||
      e.altKey ||
      e.metaKey ||
      e.ctrlKey
    ) {
      return;
    }
```

Issue #30, opened 2020-11-10
Use a more pleasant color scheme


fregante:
I haven't used `npmgraph` in a while but I'm pleasantly surprised by what it can do now!

However the military green is a little off-putting, would you consider using a [fresher color palette](https://colorhunt.co/)?

broofa:
Colors are a notoriously subjective topic.  Feel free to suggest a different palette.  The base colors are [defined here](https://github.com/broofa/npmgraph/blob/master/index.css#L5-L19).  As you can see, I found the coolors.co site to be useful.

PR #50, opened 2021-03-11
Change style slightly

fregante:
Thoughts? I tried to reduce the thick and heavy gray borders. Sections outside these screenshots haven't been affected.

Live demo: https://npmgraph-fregante.vercel.app

More can be done, but it's a start, if you'd like to entertain updates.

## Before

Screen Shot 1

## After

Screen Shot


index.css
```diff
@@ -15,11 +16,16 @@
...
+    to bottom,
+    #919191,
+    #7e7e7d
+  );
   --highlight: #ff761a;
 }
 
 html {
-  font-family: 'Inter', sans-serif;
+  font-family: system-ui, Segoe UI, Helvetica, Arial, sans-serif, Apple Color Emoji, Segoe UI Emoji;
```
fregante, line 28: This matches GitHub’s `font-family`

index.css
```diff
@@ -41,11 +47,11 @@ body {
 }
 
 .bright-hover {
-  transition:.5s;
+  transition: filter .5s;
 }
 .bright-hover:hover {
   filter: brightness(125%);
-  transition:.15s;
+  transition-duration: .15s;
```
fregante, line 54: Limit transitions to the specified property

index.css
```diff
@@ -54,17 +60,22 @@ body {
   overflow: auto;
   position: relative;
   user-select: none;
+  padding-right: var(--splitter-size);
```
fregante, line 63: Adds some space to the graph area so the splitter doesn't hide it.

index.css
```diff
@@ -94,7 +103,7 @@ label {
 summary {
   position: -webkit-sticky;
   font-size: 1.2em;
-  font-weight: bold;
+  font-weight: 600;
```
fregante, line 106: I found `bold` to be a little too thick on my side

index.css
```diff
@@ -323,7 +350,7 @@ g.edge.selected > polygon {
 #drop_target {
   background-color: #ddd;
   border: dashed 2px #888;
-  padding: 1em;
+  padding: 4em 1em;
```
fregante, line 353: Taller drop target. Ideally the whole page should be a target though, while this would just be a non-functional "invite" to drop.

js/Inspector.js
```diff
@@ -14,7 +14,7 @@ export function Fix() {
 }
 
 export function ExternalLink({ href, children, target = '_blank', className, style, ...props }) {
-  return html`<a href=${href} className="bright-hover ${className}"  target=${target} style=${{ marginRight: '8px', ...style }} ...${props}>
+  return html`<a href=${href} className="bright-hover ${className || ''}"  target=${target} style=${{ marginRight: '8px', ...style }} ...${props}>
```
fregante, line 17: To avoid `class="bright-hover undefined"`

fregante:
Also the loader now:

gif

fregante:
Ping if you're interested

broofa:
Sorry for the radio silence on this one.  I'm okay with some of this, but the change from splitter -> little knob thing is actually a regression.  I had that UI originally and it just didn't feel right to me.  To... one-off.  I opted for the more conventional splitter, even though it takes up a little more room.

fregante:
The thing is I realized that that’s what made it look ugly in #30: The “boxing” of the UI rather than the colors. Rather than having a thin clickable line I opted for a toggle with a wide square area, and it helps with the style too

fregante:
What if I restore the full-height toggle but leave it white/transparent unless hovered?

broofa (approved):
Sorry for the delay on this.  Consider this part of my "stepping back a little". :-)

fregante:
🙌

PR #52, opened 2021-03-11
Escape module name in loader

fregante:
Namespaced package names need to be escaped, notice the `%2F`:

## Before

gif

## After

2gif


broofa:
Good catch, thanks!

PR #54, opened 2021-03-13
chore: update repository metadata

fregante:
- Add server to `npm start`
- Allow file edits on GitHub.com (reverses `eol-last` setting)
- Updates the git URL in package.json
- Updates the lock to v2 (Node 15 / npm 7)


fregante:
By the way, the `gh-pages` branch is no longer necessary, so I reverted to `main`. 

Also according to the steps on `js.org` they want the CNAME to point to them already. Given that the repo was not using GitHub Pages it can stay that way already even if it's broken.

broofa:
Updated npm.broofa.com to redirect to npmgraph.js.org.  Might take a few minutes for DNS to propagate.

PR #55, opened 2021-03-13
Add file picker for package.json

fregante:
Replaces and closes #17. Reimplemented with the minimum amount of changes

Screen Shot 6

### DIFF
--- a/js/InfoPane.js
+++ b/js/InfoPane.js
@@ -38,7 +38,7 @@ export default function InfoPane() {
     if (!file) return alert('Please drop a file, not... well... whatever else it was you dropped');
 
     readFile(file);
-  }
+  };
 
   const readFile = async file => {
     const reader = new FileReader();
--- a/js/InfoPane.js
+++ b/js/InfoPane.js
@@ -13,8 +13,16 @@ export default function InfoPane() {
 
   const [recents, setRecents] = useState(getFileEntries());
 
+  // Handle file selection via input
+  const onSelect = (ev) => {
+    readFile(ev.target.files.item(0));
+
+    // Reset field
+    ev.target.value = '';
+  };
+
   // Handle file drops
-  const onDrop = async ev => {
+  const onDrop = ev => {
     ev.target.classList.remove('drag');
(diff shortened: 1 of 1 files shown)


js/InfoPane.js
```diff
@@ -13,8 +13,16 @@ export default function InfoPane() {
 
   const [recents, setRecents] = useState(getFileEntries());
 
+  // Handle file selection via input
+  const onSelect = (ev) => {
+    readFile(ev.target.files.item(0));
```
fregante, line 18: `input` already filters `.json` files, so it doesn't need further checks

js/InfoPane.js
```diff
@@ -66,16 +78,17 @@ export default function InfoPane() {
 
   return html`
     <${Pane} style=${{ display: 'flex', flexDirection: 'column' }}>
+      <input id="package-input" type="file" hidden onChange=${onSelect} accept=".json"/>
       <p>
       Enter NPM module name here <i class="material-icons">arrow_upward</i> to see the dependency graph.  Separate multiple module names with commas (e.g. <a href="?q=mocha, chalk, rimraf">"mocha, chalk, rimraf"</a>).
       </p>
-      <div id="drop_target" style="text-align: center"
+      <label for="package-input" id="drop_target" style="text-align: center"
```
fregante, line 85: The whole drop target now triggers the file selection dialog when clicked

js/InfoPane.js
```diff
@@ -66,16 +78,17 @@ export default function InfoPane() {
...
         onDrop=${onDrop}
         onDragOver=${onDragOver}
         onDragLeave=${onDragLeave}
       >
-        ... or drop a <code>package.json</code> file here
-        </div>
+        Alternatively, <button type="button">select</button> or drop a <code>package.json</code> file here
```
fregante, line 90: The button is a no-op, it's only here as a visual hint.

broofa (approved):
I'll test this once it's on `main` and I merge these and the other recent PRs with the port I'm doing to use `parcel`.

PR #61, opened 2021-04-10
chore: use Parcel to build

broofa:
Switching away from `htm`+`preact` version of the site, to use the parcel Bundler (with actual React+JSX syntax).  Currently this only works on localhost when running the Parcel dev server (`npm start`).

'Still need to figure out the process for building the static files when merging into `main`.  (something something github actions something something...?)

Also, graph rendering is broken until I can figure out how to deal with `hpcc-js/wasm`'s dynamic loading of WASM files.]


index.html
```diff
@@ -39,50 +39,52 @@
     <title>NPMGraph - Visualize NPM Module Dependencies</title>
   </head>
 
-  <body class="theme-lite"></body>
+  <body className="theme-lite"><div id='app' className='theme-lite' /></body>
```
fregante, line 42: This is a regular HTML file, className shouldn’t be used here

js/Inspector.js
```diff
@@ -112,11 +114,10 @@ export default function Inspector({ className, ...props }) {
...
       <footer className='theme-dark'>
-        NPMGraph v${ENV.appVersion} ${'\xa9'} Robert Kieffer, 2020  MIT License – <ExternalLink
-            id='github'
-            href='https://github.com/npmgraph/npmgraph'>
-          GitHub
-        </ExternalLink>
+        NPMGraph v{ENV.appVersion} {'\xa9'}, MIT License
+        (See <ExternalLink id='github' href='//github.com/npmgraph/npmgraph'>
+        GitHub
+        </ExternalLink> for details)
```
fregante, line 120: Doesn't this break on multiple lines now? I think a regular [GitHub](#) link is understandable enough

js/modifiedEvents.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/**
+ * Allow shift-click, ctrl-click, etc. in browsers to work as intended when they
+ * might otherwise be swallowed by event handlers
+ */
+export function handleModifiedClicks() {
+  document.addEventListener('click', e => {
+    // Ignore events that aren't modified link clicks
+    if (e.target.nodeName != 'A' || !eventIsModified(e)) return;
```
fregante, line 16: This breaks when clicking `<a><span>link</span></a>`. You'll have to use something like `e.target.closest?.('a')``

package.json
```diff
@@ -1,16 +1,23 @@
...
+    "since 2017-06"
+  ],
   "dependencies": {
-    "@hpcc-js/wasm": "0.3.14",
+    "@hpcc-js/wasm": "1.4.1",
+    "d3": "6.6.2",
+    "d3-graphviz": "4.0.0",
     "http-errors": "1.8.0",
-    "preact": "10.5.7"
+    "react": "17.0.2",
```
fregante, line 19: Why the switch?

package.json
```diff
@@ -1,16 +1,23 @@
...
   "version": "2.7.1",
+  "type": "module",
   "author": {
     "name": "Robert Kieffer",
     "url": "http://github.com/broofa",
     "email": "robert@broofa.com"
   },
   "license": "MIT",
+  "browserslist": [
+    "since 2017-06"
```
fregante, line 12: Why so specific?

broofa: It's the recommended solution to [this error](https://flaviocopes.com/parcel-regeneratorruntime-not-defined/), which I was running into.

package.json
broofa: I was only using `preact` because I was using `htm`.  I was only using `htm` because I didn't want a build step.  Now, with Parcel, I'd rather stick with the recommended modules unless/until there's a need

fregante: I say this because React is still larger than Preact, so if there’s no advantage in using React directly I think it’s best to stick to Preact

broofa: Let's revisit this after I get this PR landed and we have a better feel for what the production build size looks like.

broofa:
@fregante Can you take a look at https://github.com/npmgraph/npmgraph/blob/8c18634375c701b64c30b73ee50c0faaf7ccef80/.github/workflows/build.yml ?    I *think* this will work for deploying, but not at all sure this is the best/correct way to go about this.  It adds a "Push /docs" step that resets the gh-pages branch and pushes the `docs` directory to it. 

That step is currently conditional on this PR, so obviously the intent would be to change that to only happen on push actions to `ref/heads/main`.  And change the repo Settings -> Page to serve from gh-pages branch, /docs directory.

The one problem with this currently is that the parcel build breaks due to https://github.com/parcel-bundler/parcel/issues/6121 .  I'm hoping to hear back from the parcel team with a solution there.  (I'm guessing it's some babel config issue, but...???)

broofa:
@fregante Okay, I think this is really, actually, ready for review.  I apologize for big PR - this definitely turned into more than I was expecting - but it was needed.  I got rid of most of the uglier parts of this codebase.  The file organization is still a bit haphazard, but it's at least a more or less standard React+JSX+Parcel project now.  'Should be a lot easier to work with.

js/modifiedEvents.js
```diff
@@ -0,0 +1,27 @@
...
+    if (!e.target.closest('A') || !eventIsModified(e)) return;
+
+    // Stop processing this event
+    e.preventDefault();
+    e.stopPropagation();
+
+    // Perform the click on a new link (sans-event handlers)
+    const el = document.createElement('a');
+    el.href = e.target.href;
+    el.dispatchEvent(new MouseEvent('click', e));
```
fregante, line 25: I would ensure that this actually works as expected on modified clicks since this fake mouseevent would potentially be asking the browser to open a tab or download a file (alt-click). Feels like it may well be blocked in some browsers

package.json
```diff
@@ -25,21 +33,25 @@
...
-    "standard-version": "9.1.0",
-    "static-server": "*"
+    "parcel": "2.0.0-nightly.636",
+    "standard-version": "9.2.0"
   },
   "scripts": {
-    "deploy": "npm test && ssh broofa.com \"cd npm.broofa.com && ./deploy.sh\"",
-    "start": "static-server",
-    "open": "open -a \"Google Chrome\" http://localhost:9080",
+    "start": "rm -fr .parcel-cache && parcel index.html --dist-dir docs",
```
fregante, line 52: There’s a `no-cache` flag that can be used on both the build and the server

.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+  "presets": [
+      "@parcel/babel-preset-env",
+      "@babel/preset-react"
+  ],
+  "plugins": [
+    "@babel/plugin-proposal-class-properties",
```
fregante, line 7: This seems to be part of preset env, it might just need to be enabled https://babeljs.io/blog/2020/05/25/7.10.0

.babelrc
```diff
@@ -0,0 +1,10 @@
...
+    "@parcel/babel-plugin-transform-runtime"
```
fregante, line 8: I don’t think this is needed with the current browser support, it’s worth a try running without it

fregante (approved):

fregante:
Also can you publish this branch to a temporary URL? If you set up vercel instead of GitHub Pages this happens automatically for every PR, which would be useful.

js/modifiedEvents.js
broofa: Well... it works in Chrome.   In testing FF just now, however... not so much.  Crap. :-p  Good catch.

index.css
```diff
@@ -368,7 +376,8 @@ footer {
   padding: 0.5em;
   box-sizing: border-box;
   width: 32em;
-  color: var(--bg0);
+  color: var(--text);
+  background-color: var(--bg0);
```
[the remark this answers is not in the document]
broofa: My issue here is the use of a background color (`--bg0`) as a text color.  I know this is probably a bit pedantic but my approach to themes is to have semantically named vars based on where the colors are applied (e.g. background, text, highlight).  Thus, if we need a "dim" text color we should have `--dim-text` var rather than repurposing `--bg0` (which makes it difficult to tweak bg0 later on should we feel the need.)

I'll fix this up.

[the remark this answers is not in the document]
fregante: Yeah, the current theming situation is really confusing, I don't know if `--text` is gonna be dark or light. As you said, I think, there should only ever be one `.theme-*` defined globally, not per section.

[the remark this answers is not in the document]
broofa: Me writing theme code:

image

package.json
broofa: Does it do anything?  When I do `parcel serve --no-cache index.html`, it still generates the .parcel-cache directory.  😕

.babelrc
broofa: Looks like you're right.  FWIW, I was just copy/pasting [from here.](https://v2.parceljs.org/languages/babel/)

broofa:
> Also can you publish this branch to a temporary URL? If you set up vercel instead of GitHub Pages this happens automatically for every PR, which would be useful. Or I can do that in a separate PR

Assigned #62 to you

js/modifiedEvents.js
broofa: Using your filter-altered-clicks as suggested. :-D

Issue #47, opened 2021-02-02
Filter dependencies by organization


bennycode:
First of all, I really like npmgraph. It's dependency visualization is of great value!

Would it be possible to add a filter for an organization? 

I would like to be able to only see the dependencies that are from my organization. For example, when I look for `@wireapp/core`, I want to just see the dependencies that start with `@wireapp`, so it will show me `@wireapp/api-client`, `@wireapp/cryptobox`, etc. but not `tough-cookie` or any other dependency which doesn't belong to my organization.

Testing Example: https://npm.broofa.com/?q=@wireapp/core

fregante:
Related: https://github.com/broofa/npmgraph/issues/38

bennycode:
It's kind of related... I don't want to focus the organization. I want to filter everything which is not from the organization. I checked out the repo locally and applied a small adjustment which works for me:

![image](https://user-images.githubusercontent.com/469989/106871661-c2a9b800-66d2-11eb-92e7-a40360c2be19.png)

@broofa, how about making this an option in the section of the switches? I can see "Include" options and we could add "Exclude" options too. There is also a listing of "Modules", "Maintainers" & "Licenses"... How about adding a fourth for "Organizations"?

bennycode:
Here is a mockup of how I would like to filter it:

![image](https://user-images.githubusercontent.com/469989/123931731-132af600-d991-11eb-9690-1d441ad84c18.png)

imbeyondboredom:
Bumping this. Would love to be able to exclude packages that are outside of my org. I want to be able to provide wider org visibility on what level of regression needs to happen when someone updates a dependency. Or at least show the potential impact to a developer when they're about to change something in a monorepo.

PR #67, opened 2021-04-18
feat: shift-click modules to expand/collapse dependencies

broofa:
Shift-click modules in the graph to collapse/expend their dependencies.  Collapsed modules are shown as semi-transparent.

image

This puts the basic code in place for manipulating what modules/dependencies are shown in a graph, so enhancing this to allow for regex-based hiding of modules should be pretty easy.  However ... the reason I went with this approach is that it keeps the parent module in the graph.  This makes it easier to undo the collapse action (parent module is still shift-clickable in the graph), and it makes it clear where modules have been removed from the graph.

@bennycode: This isn't quite what you're asking for in #47, but it's a step in the right direction.

(@fregante Does switching to Vercel remove the need to commit the `parcel build` output (/docs/* in this diff)?    'Would love to not have that be part of the commit step.)


fregante:
The was never a need to commit the docs folder on the main branch. The original idea was to push to the gh-pages branch like the workflow I linked to did.

Anyway, yes. Vercel builds and hosts the file, so it just needs the source.

fregante:
Seems to work well and it's useful! One small issue is due to how `shift` works (in Safari only). I think it just needs a `e.preventDefault()`

gif

broofa:
> Seems to work well and it's useful! One small issue is due to how `shift` works (in Safari only). I think it just needs a `e.preventDefault()`

You mean the brief text-selection flash?  Looks like that's because Safari requires the "-webkit-" prefix for the `user-select` CSS property.  Should be fixed now.

broofa:
@fregante Okay to merge?

PR #68, opened 2021-04-18
chore: index.css -> separate .scss files

broofa:
No actual changes to styling.  Just refactoring code to take advantage of Parcel's (S)CSS support

* Switch to SCSS, using notation where appropriate
* Separate index.css into component-specific *.scss files
* `stylelint` for CSS linting (included in `npm test`)
* (... and moved favicon.* images to `images` directory to clean up root directory a little)


.stylelintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "stylelint-config-standard"
+}
```
fregante, line 3: I'd add EOL where missing

fregante (approved):

.stylelintrc.json
broofa: End-of-file newlines [are enabled](https://github.com/stylelint/stylelint-config-standard/blob/master/index.js#L78).  ... or were you referring to some other missing EOLs?

broofa:
FYI, checks are breaking because parcel part of the CSS sourcemap changes on every build (even when source doesn't change. 😢 )  Looking into it.

.stylelintrc.json
fregante: eslint doesn’t lint JSON files. This file lacks a EOL newline

PR #71, opened 2021-04-20
chore: update build to use third-party service

### LINKED ISSUE #62: Setup Vercel builds
'Took a quick look at setting this up in Vercel.  Unfortunately they don't allow organization-owned repos on personal accounts so I don't think we can do this on a free plan.  But they do [support opensource projects](https://vercel.com/support/articles/can-vercel-sponsor-my-open-source-project).

If you could take care of setting that up, that would be helpful.

Since we're going this route, It probably makes sense to have building  the `docs/` dir be a required part of making commits.  `uuid` does something like this.  The CI workflow builds the docs and fails if it results in any changed files.

fregante:
Closes #62 

I also updated Vercel’s config to use the `dist` folder.


fregante:
I just deleted the project from Netlify, so ignore those errors.

fregante:
🙌 all good now

broofa (approved):
Looks good, thanks!

image

PR #74, opened 2021-04-20
fix: empty query clears url params

### LINKED ISSUE #70: `q` parameter is always appended to the URL even if empty
1. Click https://npmgraph.js.org/
2. Click anywhere on the page to un-focus the field
3. Notice that the URL changed to https://npmgraph.js.org/?q=

broofa:
Fixes #70

### DIFF
--- a/js/App.js
+++ b/js/App.js
@@ -23,7 +23,7 @@ function Splitter({ onClick, isOpen }) {
 // Parse url query param from browser location, "q"
 function queryFromLocation() {
   const q = /q=([^&]+)/.test(location.search) && RegExp.$1;
-  return q ? decodeURIComponent(q).split(/\s*,\s*/) : [];
+  return q?.length ? decodeURIComponent(q).split(/\s*,\s*/) : [];
 }
 
 export const activity = new LoadActivity();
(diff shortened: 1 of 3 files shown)


fregante:
That looks like it could work. If you’d like to look into URLSearchParams it’d save some regexes and manual decoding. 

Something like

``` js
// set
const url = new URL(location.href)
url.searchParams.set('q', value.trim())

// get
const q = new URLSearchParams(location.search).get('q')

File: js/App.js [hunk 1]
```diff
@@ -20,10 +20,13 @@ function Splitter({ onClick, isOpen }) {
...
-// Parse url query param from browser location, "q"
+// Parse `q` query param from browser location
 function queryFromLocation() {
-  const q = /q=([^&]+)/.test(location.search) && RegExp.$1;
-  return q ? decodeURIComponent(q).split(/\s*,\s*/) : [];
+  const q = new URL(location).searchParams.get('q');
+  return decodeURIComponent(q ?? '')
+    .split(',')
+    .map(v => v.trim())
+    .filter(v => v);
```
yakov116, line 29:
```suggestion
    .filter(Boolean);
```

I think that works too

broofa: TIL... Thx!

fregante (approved):

PR #75, opened 2021-04-21
fix: improved graph rendering logic, fixes #64

### LINKED ISSUE #64: Defer loading of wasm file
If you use the `import()` function the website will be usable earlier (e.g. it can start resolving the graph) _while_ it downloads this huge dependency.

```
docs/graphvizlib.d416b260.wasm      ⚠️  1 MB
docs/index.80c6c8bb.js                  368.92 KB   
```

broofa:
I looked into using a dynamic `import()` call, but wasn't able to get that to work. I don't think Parcel is set up for `url:` style dynamic-imports.  This issue is alluded to in https://github.com/parcel-bundler/parcel/issues/4498, but 'not entirely sure that issue covers our use case.

Regardless, I'm not sure it actually matters all that much because the static import isn't actually fetching the file. All it does is yield a URL to the file (which Parcel has copied into the `dist` directory).  The `fetch()` to get the file happens in our code [here](https://github.com/npmgraph/npmgraph/blob/main/js/Graph.js#L11-L16), once the graph module is loaded.  But it's async and doesn't block the initial page load / render.

So, I don't think we can do much to optimize how/when the WASM file is loaded, but I did make a couple tweaks to improve things slightly:

* Optimize the empty-graph case to avoid awaiting the graphviz logic. (Keeps user from seeing the "Rendering..." loader  while WASM loads)
* Update the inspector React state sooner, so it's usable while the graph is rendering.

Other than that, I don't think there's much more to be done.

### DIFF
--- a/js/Graph.js
+++ b/js/Graph.js
@@ -1,6 +1,5 @@
 import React, { useState, useEffect } from 'react';
 import * as d3 from 'd3';
-
 import { store, activity, useQuery, useDepIncludes, usePane, useInspectorOpen, useModule, useGraph, useColorize } from './App';
 import { $, tagElement, report, fetchJSON } from './util';
 import { graphviz } from '@hpcc-js/wasm';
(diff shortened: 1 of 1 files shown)


fregante:
Ah I see that the package isn't actually set up in a way that it's automatically picked up by Parcel and that #64 was never a problem since the file already loaded asynchronously.

broofa:
@fregante Okay to merge?

fregante:
Always

PR #76, opened 2021-05-18
Validate HTML

fregante:
This was not really a problem, but I found few small issues (and unnecessary code) in the HTML:

https://validator.w3.org/nu/?doc=https%3A%2F%2Fnpmgraph.js.org%2F

Screen Shot

This PR fixes everything except the `|` character. It could be replaced with `%7C` too but then it wouldn't be as readable.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,9 +1,8 @@
-<!DOCTYPE html />
+<!DOCTYPE html>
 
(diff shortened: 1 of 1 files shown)


broofa (approved):
Looks good, thanks!

PR #78, opened 2021-05-21
feat: Persistence Settings (currently only Colorize)

### LINKED ISSUE #77: Feature: Hope that the settings can be persisted
Currently, the settings are not preserved, and they must be reset the next time opened, which is a bit annoying. I think it will be more intuitive if the settings can be persisted.

wenmin92:
fix #77

### DIFF
--- a/js/App.js
+++ b/js/App.js
@@ -13,7 +13,7 @@ export const useInspectorOpen = sharedStateHook(true, 'inspectorOpen');
 export const useQuery = sharedStateHook(queryFromLocation(), 'query');
 export const useModule = sharedStateHook([], 'module');
 export const useGraph = sharedStateHook([], 'graph');
-export const useColorize = sharedStateHook(false, 'colorize');
+export const useColorize = sharedStateHook(false, 'colorize', true);
 export const useDepIncludes = sharedStateHook(['dependencies'], 'depIncludes');
 export const useExcludes = sharedStateHook([], 'excludes');
 
--- a/js/sharedStateHook.js
+++ b/js/sharedStateHook.js
@@ -38,8 +38,9 @@ import { useState, useEffect } from 'react';
  * @returns {Function} React hook function
  */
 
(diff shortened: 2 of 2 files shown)


broofa (changes requested):
Great feature - this is definitely a worthwhile improvement!

A couple of comments ...

1. `localStorage` settings are shared across all windows.  If/when the user has multiple npmgraph windows open, changes to ` colorize` will apply everywhere.  It might make more sense to use `sessionStorage` (window-specific storage, but only persists for the lifetime of the window/tab.)
2. I'm not thrilled about the increased complexity of `useSharedState()`.  It's a clever place to add this behavior, but it complicates the purpose/implementation of that method.  Instead, I'd suggest switching to something like `useLocalStorage()` from [`react-use-localstorage`](https://www.npmjs.com/package/react-use-localstorage).  It has the added benefit of tracking DOM `storage` events so external changes to the property would get picked up immediately.  (Relevant because of point \#1.)

So.. would `sessionStorage` work here?   If so, is there an existing NPM module that would suffice for that?  (I've been searching around and the offerings in this space are actually pretty disappointing.  `react-use-storage` gets close, but has a few issues and looks like abandonware.)  See also https://github.com/leny/react-use-storage/issues/301

(@fregante : Thanks again for setting up Vercel.  'Made reviewing this 10X easier.)

wenmin92:
For the first comment, we can use both sessionStorage and localStorage, like this:
```javascript
value = localStorage.getItem(name) ?? value;
==> value = sessionStorage.getItem(name) ?? localStorage.getItem(name) ?? value;
...
isPersist && localStorage.setItem(name, value);
==> isPersist && (sessionStorage.setItem(name, value), localStorage.setItem(name, value));
```

For the second comment, I have no idea how to deal with shared state and session/local storage. can you do it? 

Thanks for your hard work.

broofa:
Picking this up in #79

PR #80, opened 2021-05-24
Persist colorize in sessionStorage

broofa:
Reworking #78 using `storagehooks` module to persist colorize in sessionStorage.

FWIW, `storageHooks` is my take on the "react hook for connecting component state to web storage" problemspace.  I just didn't find a module that really solved this problem properly so... yeah, wrote my own. 😢 

cc: @wenmin92

### DIFF
--- a/js/App.js
+++ b/js/App.js
@@ -1,11 +1,10 @@
-import React, { useState, useEffect } from 'react';
-import Inspector from './Inspector';
-import Graph from './Graph';
-import { LoadActivity } from './util';
-import Store from './Store';
+import React, { useEffect, useState } from 'react';
 import { Loader } from './Components';
+import Graph from './Graph';
+import Inspector from './Inspector';
 import sharedStateHook from './sharedStateHook';
-
+import Store from './Store';
+import { LoadActivity } from './util';
 import '/css/App.scss';
 
 export const usePane = sharedStateHook('info', 'pane');
@@ -13,7 +12,6 @@ export const useInspectorOpen = sharedStateHook(true, 'inspectorOpen');
 export const useQuery = sharedStateHook(queryFromLocation(), 'query');
 export const useModule = sharedStateHook([], 'module');
 export const useGraph = sharedStateHook([], 'graph');
-export const useColorize = sharedStateHook(false, 'colorize');
 export const useDepIncludes = sharedStateHook(['dependencies'], 'depIncludes');
 export const useExcludes = sharedStateHook([], 'excludes');
 
--- a/js/Graph.js
+++ b/js/Graph.js
@@ -1,8 +1,9 @@
 import { graphviz } from '@hpcc-js/wasm';
 import * as d3 from 'd3';
 import React, { useEffect, useState } from 'react';
+import { useSessionStorage } from 'storagehooks';
 import wasmUrl from 'url:@hpcc-js/wasm/dist/graphvizlib.wasm';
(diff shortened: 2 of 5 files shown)


wenmin92:
Thank you.
Maybe we can do better. No one likes to reset every time you open an app.
I think we can take care of both localStorage and sessionStorage. (Like https://github.com/npmgraph/npmgraph/pull/78#issuecomment-846639062)

broofa:
> For example, when we set a value, we put it into sessionStorage and localStorage at the same time. 

I don't think there's ever a good reason to duplicate state across `localStorage` and `sessionStorage`. You either want the state scoped to one window's session, or you want it to persist across sessions and be available on any window on the same domain.  It's really just a question of which behavior you want.

For the NPMGraph use cases, I actually think it's rare that people will have more than one window open at a time.  In the one-window use case, the only real impact of switching to `localStorage` is that this setting will persist across window sessions, which I agree would be nice.

One of the potential issues with localStorage that I want to stay cognizant of is what happens if you a user has many windows open showing large graphs.  Changes to localStorage will apply across all windows and takes effect immediately (thx `storagehooks`!), so if you select one of the NPMS.io-based metrics (e.g. maintenance score), each window will immediately hit the NPMS.io site for the score of all modules.  I'm not aware of any real performance concerns at this time - the XHR requests for that are pretty efficient - but I don't want to abuse that API unnecessarily.

I'm happy to try the localStorage option out to see how it feels.  But we should be prepared to switch back to sessionStorage if any issues emerge.

broofa:
Okay, I've switched this to `useLocalStorage`, give it a whirl.

One thing I note immediately is that this prevents you from colorizing different graphs with different metrics.  Again, the setting applies across all windows, which is probably not the behavior people will expect.

Possible solution: make colorize setting specific to the module being graphed?

Alternatively, we could put this setting in the URL hash.  E.g. `...q=module_being_graphed#colorize=overall`.   (This touches on another issue that's been nagging at me for awhile: that URLs don't capture the view settings.  So, for example, there's no way of sharing a link to a graph that includes `devDependencies` and that hides certain modules.

That actually might make more sense, now that I think about it.  😕

wenmin92:
You have considered a lot.

But I don't understand why you disagree the double storage (`localStorage` & `sessionStorage` at the same time). 

The `sessionStorage` has the highest priority, followed by `localStorage`, and finally the default value passed in when calling.

When we visit the site for the first time, `sessionStorage` will not have any value, we will take the value from `localStorage`. 
Assuming that we open another window at this time, this new window will also be initialized with the value in `localStorage`.

After that, once any new settings are made, `sessionStorage` and `localStorage` will be updated at the same time, but because `sessionStorage` has a higher priority, updating `localStorage` will not affect other windows. There is no problem that occurs when using `localStorage` alone.

fregante:
@wenmin92 what your saying makes sense, but I think it should not use sessionStorage, which is semi-permanent. I think:

1. Read localStorage only on load
2. Write when the setting is updated
3. Keep the current setting in memory

wenmin92:
> I don't see the reason to also save the current setting in the storage if it's meant to be used only buy the current window
One use case: when user refresh, memory state will be lost, but sessionStorage will not.

fregante:
`history.state`

wenmin92:
The first and last, we should init the state with the user settings. If we use URL hash, we should init it with local settings.
An example, https://babeljs.io/repl, it use both URL hash and localStorage.

PR #86, opened 2021-06-14
Darkmode, close #85

### LINKED ISSUE #85: Add dark mode / night mode
Semi-related image. We need night mode 😁 

image

broofa:
Took a stab at this since it touches on [an issue](https://github.com/npmgraph/npmgraph/pull/61#discussion_r613273528) I mentioned a while back.  It changes the look somewhat, because I wanted to avoid a proliferation of CSS vars.   'Figured we could iterate on it once the basic functionality was in place.

### DIFF
--- a/css/InfoPane.scss
+++ b/css/InfoPane.scss
@@ -1,5 +1,5 @@
 #drop_target {
-  background-color: #ddd;
+  background-color: var(--bg-dim);
   border: dashed 2px #888;
   padding: 4em 1em;
   border-radius: var(--rad_sm);
--- a/js/Graph.js
+++ b/js/Graph.js
@@ -436,6 +436,9 @@ export default function Graph(props) {
     <line stroke="rgba(0,0,0,.15)" stroke-width="6px" x1="9" x2="9" y2="12"/>
     </pattern>`;
 
+    // Remove background fill (so app theme shows through)
+    d3.select('#graph .graph polygon').remove();
+
     d3.select('#graph svg')
       .insert('defs', ':first-child')
       .html(PATTERN);
--- a/js/GraphPane.js
+++ b/js/GraphPane.js
@@ -1,10 +1,10 @@
 import * as d3 from 'd3';
 import React, { useEffect, useRef } from 'react';
 import { useColorize, useDepIncludes, useExcludes } from './App';
-import { Pane, Section, Tags, Tag } from './Inspector';
-import { simplur } from './util';
 import { Toggle } from './Components';
 import { hslFor } from './Graph';
+import { Pane, Section, Tag, Tags } from './Inspector';
+import { simplur } from './util';
 import '/css/GraphPane.scss';
 
 function DepInclude({ type, ...props }) {
@@ -164,7 +164,7 @@ export default function GraphPane({ graph, ...props }) {
         }
       </Tags>
 
-      <div style={{ fontSize: '90%', color: 'var(--text-dim)', marginTop: '1em' }}>
+      <div style={{ fontSize: '90%', color: 'var(fg-dim)', marginTop: '1em' }}>
         (Shift-click modules in graph to expand/collapse)
       </div>
     </Section>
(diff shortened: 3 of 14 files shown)


fregante:
Good to see the variables cleanup! The colors don’t match the previous ones on the regular version though

broofa:
Fixed(ish).  Colors aren't an exact match because I'm just inverting the fg/bg colors for the tab bar.  This avoids creating more CSS vars just for this purpose, which I'd like to avoid if possible.  It makes the lights and darks there a bit more saturated but I don't see that as a bad thing, per se.

(Note: This also includes some some tweaks to the dependency and colorize controls so they're not quite so janky.)

Before:
image

After:
image

After (dark mode):
image

fregante:
I'm not sure why the colors aren't an exact match in the regular version, neither version is good-looking. There are borders where there weren't before. and it's white where it wasn't.

If you want to fix the CSS variables issue using the correct colors, maybe I can look into the dark mode later.

js/index.js
```diff
@@ -13,7 +19,7 @@ window.addEventListener('error', err => {
   Flash(err.message);
 });
 
-window.addEventListener('unhandledrejection', err => {
+window.addEventListener('nhandledrejection', err => {
```
fregante, line 22: typo

js/GraphPane.js
```diff
@@ -164,7 +161,7 @@ export default function GraphPane({ graph, ...props }) {
         }
       </Tags>
 
-      <div style={{ fontSize: '90%', color: 'var(--text-dim)', marginTop: '1em' }}>
+      <div style={{ fontSize: '90%', color: 'var(fg-dim)', marginTop: '1em' }}>
```
fregante, line 164: typo

js/GraphPane.js
```diff
@@ -120,21 +120,18 @@ export default function GraphPane({ graph, ...props }) {
...
-    Include:
     <DepInclude type='dependencies' />
     <DepInclude type='devDependencies' />
     <DepInclude type='peerDependencies' />
 
     <label>
-      Colorize by:
-      <select value={colorize ?? ''} onChange={e => setColorize(e.target.value)}>
+      <select value={colorize ?? ''} onChange={e => setColorize(e.target.value)} style={{ borderRadius: '0.5em', padding: 4, marginTop: '0.5em' }}>
+        <option value=''>Colorize By ...</option>
```
fregante, line 129: This change is unrelated and also it means we'll have a dropdown with the text "NPMS.io overall score" without a label describing what that is

js/Graph.js
```diff
@@ -436,6 +436,9 @@ export default function Graph(props) {
     <line stroke="rgba(0,0,0,.15)" stroke-width="6px" x1="9" x2="9" y2="12"/>
     </pattern>`;
 
+    // Remove background fill (so app theme shows through)
+    d3.select('#graph .graph polygon').remove();
```
fregante, line 440: I think we should keep it so it can be exported (right?), but color it via CSS

css/InfoPane.scss
```diff
@@ -1,13 +1,13 @@
...
-  background-color: #ddd;
+  background-color: var(--bg-dim);
   border: dashed 2px #888;
   padding: 4em 1em;
   border-radius: var(--rad_sm);
 }
 
 #drop_target.drag {
-  background-color: #ccf;
-  border-color: #68a;
```
fregante, line 10: Do we need more color variables?

js/Graph.js
broofa: I don't think so.  This is just removing the background rectangle that `graphviz` adds (and fills with white). It doesn't serve any purpose when printed, and otherwise just obscures the background color of the page, which we want to show through.

js/GraphPane.js
broofa: The initial value is "Colorize By ...", so users will see that when the page loads.  They'll also see it as the top option when the drop-down is opened.

fregante: I know what you mean, but this remains a label-less field. https://www.nngroup.com/articles/form-design-placeholders/

PR #91, opened 2021-07-06
Add dark mode

fregante:
This PR brings minimal changes just to add support for dark mode without affecting other parts of the code and without changing the light mode (which only has minimal changes).

The sidebar is slightly tinted red/orange, but it could be easily just brought back to a gray. Further theme improvements could be made later, but at least now we can have a workable theme without large swaths of white.

Screen Shot 10

Screen Shot 11

### DIFF
--- a/css/Graph.scss
+++ b/css/Graph.scss
@@ -22,7 +22,7 @@
     width: 1.3em;
     padding: 0;
     margin: 0;
-    color: #333;
+    color: var(--text);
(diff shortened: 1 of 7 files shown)


fregante:
The second commit colors the links in the sidebar and ensures text in the graphs stays readable:

Screen Shot 12Screen Shot 13

fregante:
Merging for now, improvements can be made on top of this.

PR #108, opened 2021-11-08
fix: graph nodes use theme colors

broofa:
Graph nodes were always rendering in "light" colors, which is annoying.  Now they follow user's system theme...

Before:
image

After:
image


fregante:
"Lint fix" needs to be run as well

PR #109, opened 2021-11-08
fix: remove theme-lite/theme-dark abstractions

broofa:
This should not result in any changes to the UI.  Just getting rid of the last vestiges of the theme-dark / theme-light that I had originally used for theming the UI.  The only real change required for this was to create the `--bg-tabs` var to abstract out the background color used by the tabs, splitter, and loader.

Everything else here is just cleanup that flows from the above.

### DIFF
--- a/css/App.scss
+++ b/css/App.scss
@@ -22,8 +22,8 @@
   display: flex;
   box-sizing: border-box;
   height: 24pt;
-  color: #fff;
-  background: var(--bg-tabs) var(--bg-gradient);
+  color: var(--tabs-text);
+  background: var(--tabs-bg) var(--bg-gradient);
   border-top-left-radius: var(--rad_lg);
   border-bottom-left-radius: var(--rad_lg);
   padding-left: 10px;
--- a/css/App.scss
+++ b/css/App.scss
@@ -22,7 +22,7 @@
   display: flex;
   box-sizing: border-box;
   height: 24pt;
-  color: var(--text);
+  color: #fff;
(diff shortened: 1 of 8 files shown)


fregante:
Great! I only see one issue: the header text is black in the light theme. 

Lint is failing but that's autofixable

css/Inspector.scss
```diff
@@ -74,7 +74,7 @@ summary {
   flex-basis: auto;
   flex-shrink: 0;
   display: flex;
-  background: var(--bg0) var(--bg-gradient);
+  background: var(--bg-tabs) var(--bg-gradient);
   color: var(--text);
```
fregante, line 78:
```suggestion
  color: #fff;
```

css/App.scss
```diff
@@ -23,7 +23,7 @@
   box-sizing: border-box;
   height: 24pt;
   color: var(--text);
```
fregante, line 25:
```suggestion
  color: #fff;
```

fregante (approved):
This fixes it. A full white also increases the contrast ever so slightly

fregante (approved):
LGTM. 

begif


Let's also add a `color-scheme: light dark;` CSS on `:root`

https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

fregante (approved):

css/index.scss
```diff
@@ -1,4 +1,6 @@
 :root {
+  color-scheme: light dark;
```
fregante, line 2: Courtesy of `color-theme`

begif

PR #111, opened 2021-11-09
fix: minor theme color issues

broofa:
* arrowheads were not filled in in dark mode (see below)
* remove background `polygon` in SVG to allow page background to show through
* remove `html` `background` CSS so the user-agent background is used
* remove the `[fill=...]` and `[stroke=...]` CSS attribute-hacks for tweaking SVG fill and stroke colors, as they no longer apply to anything.

Before:
CleanShot 2021-11-09 at 06 56 23@2x

After:
CleanShot 2021-11-09 at 06 55 19@2x

### DIFF
--- a/css/Graph.scss
+++ b/css/Graph.scss
@@ -89,13 +89,22 @@ g.node {
   }
 }
 
-g.edge.selected {
-  > path {
-    stroke-width: 2;
-    stroke: var(--highlight);
+g.edge {
+  > path { // stylelint-disable-line no-descending-specificity
+    stroke: var(--text);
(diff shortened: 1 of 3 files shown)


fregante:
> * remove `html` `background` CSS so the user-agent background is used

Not sure it's a good idea, the background is courtesy of `color-scheme`, which may or may not be supported. Also in Safari it uses a different tint than our dark color (our darks are a little orange) and with less contrast, so it looks a bit weird


## `main`



## this PR

fregante (approved):

PR #113, opened 2021-11-15
Port to Typescript, Prettier for linting

broofa:
'Been meaning to switch to TS for a while.  This provides some much-needed guard rails around the data structures being passed around.   E.g. the package / module structures we get from the npmjs repository.   While this isn't exactly rigorous (I only did a quick pass at defining the type structures there) it's a start, and provides something that can be iterated on.

In the midst of this work, I found myself struggling to get VSCode + typescript + parcel to all play nice with code linting and errors.   Using Prettier and the Prettier plugin in VSCode for formatting go that under control.  I know it makes for a lot of code churn, but the end result should be easier to maintain.

There's no functional changes here (other than 1-2 small existing bugs that Typescript identified)


.stylelintrc.json
```diff
@@ -1,3 +0,0 @@
-{
-  "extends": "stylelint-config-standard"
-}
```
fregante, line 3: No more stylelint?

package.json
```diff
@@ -42,26 +42,24 @@
...
     "standard-version": "9.3.0",
-    "stylelint": "13.13.1",
-    "stylelint-config-standard": "22.0.0"
+    "typescript": "4.4.4"
   },
   "scripts": {
     "build": "parcel build index.html",
     "release": "standard-version",
     "start": "parcel serve index.html",
-    "test": "eslint . && stylelint css",
```
fregante, line 64: Eslint is not the same as prettier though. Prettier is just formatting

fregante (approved):

.stylelintrc.json
broofa: `prettier` [styles all the thingz](https://prettier.io/docs/en/index.html)

package.json
broofa: Dooh!  Right you are.  Will fix.

.stylelintrc.json
fregante: Style lint is *also* not just formatting 😃 

It’s ok though

package.json
broofa: Holy fuck, trying to get vscode, eslint, prettier, typescript, react & JSX to all play nice ...
image

broofa: Well, that wasn't any trouble at all 🙄

.stylelintrc.json
broofa: If I have to deal with another permutation of tooling at this point I'm gonna start pulling the wings off of butterflies.  🤯

broofa: Made a brief attempt at adding styleint back in (using `stylelint-prettier` and `stylelint-config-prettier`), but gave up when it started bitching about "When linting something other than CSS, you should install an appropriate syntax, e.g. "postcss-scss", and use the "customSyntax" option" after I'd told it to not lint the `.tsx` files.

I'm going to call this good for now.

PR #114, opened 2021-11-22
fix: center small graphs

### LINKED ISSUE #110: Be smarter about inital graph zoom & positioning
Small graphs should be centered  when initially displayed.  And probably zoomed in slightly (1.5-2x?) just to fill the space a little better.

CleanShot 2021-11-09 at 06 08 49@2x

broofa:
Center-aligns graphs in 1:1 mode.  Fixes #110.

### DIFF
--- a/css/Graph.scss
+++ b/css/Graph.scss
@@ -1,4 +1,10 @@
 #graph {
+  &.centered {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+
   flex-grow: 1;
   height: 100vh;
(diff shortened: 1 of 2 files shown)


fregante:
Last time I tried this, content got clipped (I think large trees)

broofa:
> Last time I tried this, content got clipped (I think large trees)

Are you referring to this PR or something else you/I tried previously?

This diff should only change the behavior in the specific case of 1:1 zoom where the graph is known to not need scrollbars.

PR #115, opened 2021-12-10
feat: opensearch support

broofa:


broofa:
Ugh.  Thought this was going to be simple, but it looks like OpenSearch documents aren't picked up unless the user has "done a search" already.  But the mechanics of what "search" means are [very narrowly defined](https://www.chromium.org/tab-to-search/).  Getting this to work will involve reworking the markup to have an actual form with traditional submit-button mechanics rather than handling this in script as we do currently.

PR #116, opened 2021-12-10
fix: broken request error handling

broofa:
One of my more recent PRs (too lazy to track down which one) introduced a bug that caused rendering to break if one of the modules was not found in the NPM repo.  This was due to failed repo requests *not* resulting in a stub module, which led to graphs that looked like this:
CleanShot 2021-12-10 at 09 19 39@2x

This fix returns the prior (correct) behavior, and also improves error messaging shown in the module pain in the event there is an error. E.g.:

CleanShot 2021-12-10 at 09 21 52@2x

To test: Show any graph that has a non-existent module.  E.g. https://npmgraph.js.org/?q=cgm-remote-monitor

### DIFF
--- a/js/Store.ts
+++ b/js/Store.ts
@@ -2,7 +2,7 @@ import doesSatisfySemver from 'semver/functions/satisfies';
 import isSemverValid from 'semver/functions/valid';
 import Module, { moduleKey } from './Module';
 import { ModuleInfo } from './types';
-import { fetchJSON, LoadActivity, report } from './util';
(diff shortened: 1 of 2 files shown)


broofa:
@fregante : BTW, I [belatedly] noticed this while [finally] taking a long-overdue at the BugSnag reports.  It occurs to me it might make sense to give you access to that bugsnag project.  Is that something you'd be interested in?

fregante:
> Is that something you'd be interested in?

I'm ok without it 😃 

I'm here just to make it look pretty and usable 😁

PR #119, opened 2022-01-13
add filtering by regex

Hades32:
This is very handy to e.g. show only dependencies of your own NPM scope

### DIFF
--- a/js/App.tsx
+++ b/js/App.tsx
@@ -14,6 +14,7 @@ export const useQuery = sharedStateHook(queryFromLocation(), 'query');
 export const useModule = sharedStateHook([], 'module');
 export const useGraph = sharedStateHook(null as GraphState, 'graph');
 export const useColorize = sharedStateHook('', 'colorize');
+export const useFilterPattern = sharedStateHook('', 'filterPattern');
 export const useIncludeDev = sharedStateHook(false, 'includeDev');
 export const useExcludes = sharedStateHook([], 'excludes');
 
--- a/js/Graph.tsx
+++ b/js/Graph.tsx
@@ -9,6 +9,7 @@ import {
   store,
   useColorize,
   useExcludes,
+  useFilterPattern,
   useGraph,
   useIncludeDev,
   useInspectorOpen,
@@ -411,6 +412,7 @@ export default function Graph() {
   const [graph, setGraph] = useGraph();
   const [excludes, setExcludes] = useExcludes();
   const [colorize] = useColorize();
+  const [filterPattern] = useFilterPattern();
 
   const [zoom, setZoom] = useState(0);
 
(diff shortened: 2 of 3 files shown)


broofa:
Thanks for the PR.  I like where this is headed.  That said, something about this feels "off".  I think it's that under the hood this is using the same expand/collapse behavior of the (ill-named) `excludes`  state.

E.g. If I filter on "request" in [the `request` graph](https://npmgraph-git-fork-hades32-pattern-filter-npmgraph.vercel.app/?q=request), I get `request` plus a bunch of other (ghosted) modules.  That's not really how I would expect a filter to work.  I'd expect it to show *just* the `request` module.

Also, if this is just going to emulate how expand/collapse works, it should probably use the `excludes` state under the hood so users can leverage the same, related functionality (such as shift-clicking a module to toggle it's visibility).

I'm open to suggestions on how to address this.  I'm sure there's a way to improve the UI that meets your use case, I'd just like to make sure it's done in a way that avoids confusion.

E.g. maybe allowing users to filter by regex in the "Modules" section, then expand/collapse matching modules?  Something like this...
CleanShot 2022-01-13 at 10 49 41@2x

(Note: Just free-thinking here.  I'm not sure this approach is 💯 solid, either.)

Hades32:
Yes, it seems I was confused by the name "excludes", I now can see the issue with the ghosted modules, that's not how I intended it, I wanted it just as you described: Only show things that mach the pattern.

In that case, do you think the UI is fine, if I change the behavior to completely hide non-matched modules?

broofa:
> In that case, do you think the UI is fine, if I change the behavior to completely hide non-matched modules?

Can you help me better understand your use case here? Maybe share some examples of graphs you're looking at, and the kinds of graphs you'd like to end up with after applying a filter?

Asking because I think this feature will be confusing if it's not properly thought through.  For example, if I'm looking at the [`webpack` graph](https://npmgraph-git-fork-hades32-pattern-filter-npmgraph.vercel.app/?q=webpack) and I filter on "@webassembly" what I get isn't really what I expect.  What I get is a single ghosted `webpack` module ("'webpack' doesn't match the filter, so it gets collapsed").   And if the logic is changed to hide all non-matching modules, this probably just turns into an empty graph.

What I would expect, however, is something more like `npm ls (some module)`... a graph that shows me how the filter-matching modules (@webassembly*) connect to the root module, with any superfluous modules hidden.  Is that what you're after, or did you have something else in mind?

Hades32:
So my main use-case is to look at a package and only see the graph of packages of my own scope (i.e. `@myorg/...`), so I can understand the interdependencies better.

I agree, the root module should always be visible. What we want is basically the graph that you currently get when you use `^webpack\b|@webassembly` but without the ghosted modules.

broofa:
Closing this out.  While I do think there's some utility in being able to filter the graph, I think it's going to require a more nuanced approach.  Something that allows for a variety of filtering mechanisms (e.g. filter by depth, by size, by namespace, etc.)... and that's not a priority for me at the moment.

My apologies for dragging my feet on this.  I do appreciate the contribution and feedback!

PR #129, opened 2022-04-13
Switch to using registry.npmjs.org, update libs

broofa:
In https://github.com/npmjs-cf/meta/issues/7, @terinjokes mentions that `registry.npmjs.org` is now served with CORS headers.  Upon further investigation, it looks like it's now served through cloudflare, which is pretty much the whole point of the `.cf` mirror.  Given this, there doesn't appear to be much reason for us to continue using the `.cf` mirror. 

I also ran `npx npm-check-updates -u` to update all our dependencies.  This was "mostly painless".  (`react` and `react-dom` required a couple minor tweaks due to the change from 17 -> 18).

### DIFF
--- a/index.html
+++ b/index.html
@@ -80,7 +80,6 @@
     background-color: var(--bg0);
     left: 0;
     top: 0;
-    bottom: 0;
     right: 35%;
     border-radius: 2rem 2rem;
     color: var(--text);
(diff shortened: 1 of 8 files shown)


terinjokes:
Per npm's comments, it seems this is currently a soft launch, and I may have jumped the gun in suggesting it.

broofa:
> Per npm's comments, it seems this is currently a soft launch, and I may have jumped the gun in suggesting it.

@terinjokes What comments?  (link?)

terinjokes:
https://github.com/npm/feedback/discussions/117#discussioncomment-2561212

fregante:
> I also ran npx npm-check-updates -u to update all our dependencies.

It's always best to submit PRs separately so the changes are easier to review and merge

fregante:
npmgraph is currently broken so I'm merging this PR. Feel free to add a secondary/fallback `fetch` to the previous endpoint to make sure it works regardless of npm’s decision on CORS or any A/B testing.

broofa:
> npmgraph is currently broken so I'm merging this PR. Feel free to add a secondary/fallback `fetch` to the previous endpoint to make sure it works regardless of npm’s decision on CORS or any A/B testing.

We'll have to keep an eye on this over the coming weeks, as it seems like NPM is still in the process of rolling this out.  E.g. Not all requests are responding with CORS headers...

CleanShot 2022-04-14 at 06 29 18@2x

PR #133, opened 2022-10-27
Use lowercase for npm

friederbluemle:
Just a minor capitalization fix: `NPM` -> `npm`

npm should never be capitalized:
https://github.com/npm/cli#is-it-npm-or-npm-or-npm

Also, npms.io does not use uppercase characters:
https://npms.io/about

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# NPMGraph
+# npmgraph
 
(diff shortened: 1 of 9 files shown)


broofa:
Can't argue with sound logic.  Thanks for the contribution!

PR #134, opened 2022-10-29
add fregante to FUNDING

broofa:
@fregante I assume you're okay with this?  (Went on a bit of a binge last night adding this to various repos and forgot to include you here.)


fregante (approved):
Thank you! Appreciated

broofa:
@fregante Sorry for all the notifications about pushes and vercel builds.  I inadvertently broke things by switching from package-lock.json to yarn.lock.

On a slightly different note, this appears to have surfaced a bug in the npmjs.com website that may be of interest.   I doubt it matters all that much since this project isn't something people are going to be looking for on npm directly but figured I'd let you know I'm aware of the issue: https://github.com/npm/feedback/discussions/802

PR #155, opened 2023-09-25
fix: ability to remove uploaded files, fixes #118

### LINKED ISSUE #118: Allow forgetting the cached data for uploaded package.json
When selecting a package.json file, the name+version will probably not change when you edit locally (as the file probably does not have a meaningful version anyway if it is not a published package). As the npmgraph tool uses sessionStorage as the cache, it currently ignores any change I do in my package.json when I reupload, unless I close my browser or I manually delete the cache through the dev tools.

image

Maybe the best solution would be that a new selected file always replaces the cache. But otherwise, I would be fine with at least a button to delete a cached file from this "recent files" list.

broofa:


broofa:
@fregante Hey, I hope it's okay that I've been flying kind of fast and loose with the commits and PRs here lately.  I had some [rare] spare time and [even rarer] energy to work on this over the last week/weekend.  So I've just been pushing as quickly as possible while this spurt of energy lasts.  :-)

Let me know if you see anything concerning or have questions.

fregante:
No worries, I'm just glad to see some activity here

PR #156, opened 2023-09-26
feat: accept URLs to package.json files in addition to NPM names

broofa:
Allows you to provide a url to a `package.json` file in the search field.  This requires a bit more room in the search field, so I've had to tweak the UI a bit.

For example: https://npmgraph-git-uploadurl-npmgraph.vercel.app/?q=https%3A%2F%2Fgithub.com%2Fbroofa%2Fgmaul%2Fblob%2Fmain%2Fpackage.json

image

This treats the urls like module names, so you can actually do comma-separated URLs... which would be useful for comparing dependencies between (for example) different tags or branches of a project, but for the fact `ModuleCache` keys off the `name` and `version` in the package.json file, so that only works if those fields differ in the fetched package.json files.

Note: This also (somewhat suprisingly) will work for private github repos, as long as you use the direct link to the "Raw" file (which includes an access `token`)

Thoughts?


fregante:
Can you restore this border style on the field? Without the search icon it's ok:

Screenshot 2


This is what it looks like currently in Safari (it's the default "3D inset" `border-style`)

Screenshot 1

broofa:
Willl do.  Let me get this other PR merged and then I'll take care of this.  (And if I forget, please don't hesitate to ping me here. :-) )

PR #157, opened 2023-09-27
feat: add uploaded file sharing support

broofa:
Adds support for "uploading" package.json contents via the url `hash`.

For URLs containing a `hash` of the form `...#package_json=<URI-encoded content>`, the app will now treat the `package_json` value as it would an uploaded package.json file.

Important:  Upon "upload", the `package_json` portion of the hash will be immediately cleared.

This PR also adds a "share" button in the top-right corner of the inspector:
image

Clicking this button will copy the appropriate link to the pasteboard, _including when the graph is for a locally-uploaded file.   I.e. **This is how you get links with the `package_json` `hash` prefilled**.

For example, here's one for an old "message_tool" project of mine:
https://npmgraph-git-uploadhash-npmgraph.vercel.app/?q=local%3Amessage_tool%401.0.0#package_json=%7B%22name%22%3A%22message_tool%22%2C%22version%22%3A%221.0.0%22%2C%22description%22%3A%22Tool+for+inspecting+the+iOS+Message+app+information%22%2C%22main%22%3A%22app.js%22%2C%22dependencies%22%3A%7B%22autolinker%22%3A%221.6.0%22%2C%22colors%22%3A%221.1.2%22%2C%22escape-html%22%3A%221.0.3%22%2C%22express%22%3A%224.16.2%22%2C%22morgan%22%3A%221.9.0%22%2C%22sequelize%22%3A%224.22.0%22%2C%22sqlite3%22%3A%224.1.1%22%2C%22uniqolor%22%3A%220.9.0%22%7D%2C%22devDependencies%22%3A%7B%7D%2C%22scripts%22%3A%7B%22test%22%3A%22echo+%5C%22Error%3A+no+test+specified%5C%22+%26%26+exit+1%22%2C%22dbm%22%3A%22sqlite3+%24HOME%2FLibrary%2FMessages%2Fchat.db%22%2C%22dbc%22%3A%22sqlite3+%24HOME%2FLibrary%2FApplication%5C%5C+Support%2FAddressBook%2FAddressBook-v22.abcddb%22%7D%2C%22author%22%3A%22Robert+Kieffer%22%2C%22license%22%3A%22MIT%22%2C%22_local%22%3Atrue%2C%22maintainers%22%3A%5B%5D%7D


fregante:
Sounds great! I have a couple of ideas:

- strip the JSON file to the bare minimum before sharing, so `name,dependencies,devDependencies,etc`
- automatically include the contents in the location.href, no need to duplicate the native Share button

fregante:
Also:

- if you're concerned about URL size and noise, I suggest compressing the string. I usually use https://www.npmjs.com/package/lz-string

broofa:
Hmm... good points.

> strip the JSON file to the bare minimum before sharing, so name,dependencies,devDependencies,etc

Good idea.  Both for size and to mitigate the risk of leaking sensitive info

> automatically include the contents in the location.href, no need to duplicate the native Share button

Hmm... yeah, I probably overcomplicated this.  Simply attaching the content as part of the hash 

> if you're concerned about URL size and noise, I suggest compressing the string

I'm going to hold off on this for now.  Compressing the string makes it more difficult for users to generate their own URLs.

fregante:
> generate their own URLs.

You can still support both. `if !startsWith('{'), decompress()`

broofa:
@fregante I've reworked this so upploaded packages are now placed in the `location#hash`.  This feels pretty good.  The one question I have is whether or not I should be using `location.search` instead.  Since I'm treating `hash` as a parameterized field in exactly the same way `search` is, the two are interchangeable from an implementation standpoint. Thoughts?

One nice benefit: I was able to remove all the localStorage-related logic and UI, since the browser history now effectively acts as the storage.

components/InfoPane/QueryInput.tsx
fregante, line 1: Nit: The field should probably not include the filename:

Screenshot 2

because when the `name` is missing, this isn't useful:

Screenshot

Live demo: https://npmgraph-c4c7vmywv-npmgraph.vercel.app/?q=%28upload%29%402023.241.63122#packages=%5B%7B%22dependencies%22%3A%7B%22%40types%2Fchrome%22%3A%22%5E0.0.197%22%7D%7D%5D

fregante:
> I was able to remove all the localStorage-related logic and UI

That's great! I was never a fan of that

fregante (approved):
Looks great! There's only one nit I'm not sure about. Feel free to ignore

components/InfoPane/QueryInput.tsx
broofa: Can you open a new issue for this?

Also, I'm not sure I entirely understand the issue you're describing.  Is it simply that users shouldn't see the (admittedly very-contrived) "(upload)@...."-type names in the search field?

broofa: @fregante See https://github.com/npmgraph/npmgraph/pull/162 .  Let me know if it fixes the problem.

PR #164, opened 2023-10-01
Style fixes 

fregante:
I might add here some style improvements

### DIFF
--- a/components/InfoPane/QueryInput.scss
+++ b/components/InfoPane/QueryInput.scss
@@ -1,5 +1,4 @@
 #search-field {
-  margin-left: 1em;
   outline: none;
   border: solid 1px var(--text-dim);
   border-radius: 15px;
--- a/components/Inspector.scss
+++ b/components/Inspector.scss
@@ -95,20 +95,6 @@ summary {
       background: var(--highlight);
     }
   }
-
-  #search-field {
-    margin-left: 1em;
-    outline: none;
-    border: solid 1px white;
-    border-radius: 15px;
-    padding: 0.3em 0.7em;
-    font: inherit;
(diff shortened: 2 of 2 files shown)


components/InfoPane/QueryInput.scss
fregante, line 1: Is this rule now unused? Should drop it? 

https://github.com/npmgraph/npmgraph/blob/9745a479bb202324e2b50096aa05627a9b685856/components/Inspector.scss#L73

components/InfoPane/QueryInput.scss
```diff
@@ -1,5 +1,4 @@
 #search-field {
-  margin-left: 1em;
```
fregante, line 2: ## Before

Screenshot


## After

Screenshot 1

components/InfoPane/QueryInput.scss
broofa: It's still used.  `#tabs` is the bar at the top of the inspector.

fregante: Tabs yes, but where is the input?

components/InfoPane/QueryInput.scss
broofa: looks good!

components/InfoPane/QueryInput.scss
broofa: I think I see what you're referring to.  The `#tabs #search-field` in Inspector.scss rule no longer applies, so I've removed that.   Is that what you meant?

broofa (approved):

broofa:
'Merged this since it'll be helpful in testing another branch I'm working on (coming soon).

PR #167, opened 2023-10-05
feat: add "About" pane w/ info and sponsor buttons

broofa:
@fregante 'Found myself thinking about how to better promote the fact people can sponsor us without being obnoxious about it and came up with this.

Adds an "About" tab to the inspector, that will light up with a badge showing the number of commits "new" commits that have been made to the project.  ("New" meaning, since the user last looked at the about tab.)   The badge resets once they've viewed the tab.

CleanShot 2023-10-05 at 12 00 20@2x

The about pane is intentionally barebones.  A link to this repo, a couple sponsor buttons for you and I, and then a list of the commits.

CleanShot 2023-10-05 at 12 00 46@2x

Edit to add: If you want to reset the notification badge (e.g. to play around with it in development), delete the `lastVisit` key in `localStorage` and reload the page.


fregante:
I don’t think it's worth unliking the changes when you can just link to the commits page on GitHub. Also that "notifications" badge is distracting. 

As for the sponsor part it's up to you, but you can remove my sponsor link from there since this is 99% your work 👌

broofa:
@fregante : Thanks for the feedback.  I've made the badge on the about tab less visually intrusive. I've also removed your sponsor button.  Let me know if you change your mind about that.

fregante:
The number was not an issue, recent changes is not something people need to take action on so the fake notification badge is distracting and spammy (as many websites use it to entice you to click on some button). My stance on the whole release notes still stands

PR #170, opened 2023-10-12
feat: last-modified date colorizer

broofa:
Adds ability to colorize by "last modified date".

Example: https://npmgraph-git-datecolorizer-npmgraph.vercel.app/?q=express#color=modified

I'm a little unsure whether to merge this.   The "Last modified date" is last time [the package record was modified in the NPM registry](https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md#abbreviated-metadata-format).  That'll generally be the date of the last version that was published, but could also be something unrelated (e.g. when a version was unpublished, or an owner changed or... whatever.)  I'm concerned this could be more confusing or misleading than actually helpful.

Thoughts?


broofa:
Closing unmerged.  This is going to be more confusing than it will be useful, I think.

PR #171, opened 2023-10-12
feat: selection shows full up/downstream dependencies (fixes #144, fixes #72)

### LINKED ISSUE #72: Feature: Module selection should highlight whole dependency tree, and call out any dependencies unique to the selected module.
One question I often have about a dependency graph is, "If I add/remove [module X], what dependencies _actually_ get added / removed from the graph."  I don't know of any tools that answer this particularly well, but I think we're well positioned to do so.

For example, we could enhance the current module selection to highlight the whole descendent dependency tree, rather than just immediate child dependencies.  And, further highlight which of those dependencies are exclusive to the selected module. (i.e. which ones would actually get removed if the selected module is removed.)

#67 sort of does this in that when you collapse a module, any dependencies it shares with other modules will remain in the graph - but it's hard to see due to how the graph gets re-laid out.

### LINKED ISSUE #144: Graph selection improvements
When a module is selected in the graph currently, all directly-connected edges (dependency inlets and outlets) are highlighted.  While this does make the module a bit easier to spot in large graphs, it's not particularly useful otherwise.  Instead, we should **highlight the paths that connect the root module to the selected module**.

Furthermore, we should make it possible for users to share URLs that include the current selection.  To that end:
* Add support for including the name of the currently selected module as a `hash` on the URL.
* Update the hash in URL when the selection changes
* When the page loads, apply any selection found in the `hash`

The end result should look something like this (selection colors TBD):
CleanShot 2023-08-30 at 16 04 02

broofa:
Changes selection in the graph diagram as follows:
* Selected module highlighted in orange (as before)
* All downstream dependencies highlighted in orange (not just immediate dependencies)
* All upstream dependencies highlighted in blue
* Any edges or nodes outside of the selected modules' dependency graph are dimmed (opacity=0.35)

@fregante Any feedback you have on behavior or styling would be great.   I've been wanting to implement this selection behavior for a long time, as it feels a lot more intuitive to me so I'm hoping you think it's an improvement. :-)

CleanShot 2023-10-12 at 11 20 21@2x

### DIFF
--- a/components/GraphDiagram/GraphDiagram.scss
+++ b/components/GraphDiagram/GraphDiagram.scss
@@ -86,9 +86,27 @@ g.node {
     opacity: 0.5;
   }
 
-  &.selected > path {
-    stroke-width: 3;
-    stroke: var(--highlight);
+  &.selected {
+    & > path {
+      stroke-width: 3;
+      stroke: var(--highlight);
+    }
+  }
+
+  &.upstream {
+    & > path {
+      stroke: var(--bg-blue);
+    }
+  }
+
+  &.downstream {
+    & > path {
+      stroke: var(--bg-orange);
+    }
+  }
+
+  &.unselected {
+    opacity: 0.35;
   }
 
   &.warning > path {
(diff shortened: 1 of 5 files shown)


fregante (approved):
Agreed 👍

PR #176, opened 2023-10-16
Move splitter to the right

fregante:
Placing it there was not necessary. I think that keeping it stable in the corner makes more sense

### DIFF
--- a/components/App/App.scss
+++ b/components/App/App.scss
@@ -18,21 +18,21 @@
 }
 
 #splitter {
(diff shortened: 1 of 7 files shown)


fregante:
It might be time to rename the splitter to just "toggle". It was a splitter only when it was a tiny line between the content and the sidebar, now it's just a toggle button.

broofa:
Good call!  Nice to have you back in the code. :-)

PR #179, opened 2023-10-19
feat: display module version outdated-ness, closes #175

### LINKED ISSUE #138: Display outdated / latest versions
It would be interesting to find out which packages are in need of being updated.

### LINKED ISSUE #175: Show `latest` module (and other `dist` tags?) in Module Pane
Pretty much what #138 was asking for, but I closed that out after adding the OutdatedColorizer.  So this issue is specifically about surfacing the `latest` version(s) of a module in the UI pane.

See that issue suggested UI.

cc: @jimmywarting

broofa:
Adds a component in the module pane that shows how the module version compares to `latest`.

NOTE:  I disable this component when the user has "Colorize by:" active and set to something other than "Outdated Level".  I'm not 100% sure that's the right thing to do, but  the different color semantics felt a bit jarring.

CleanShot 2023-10-18 at 21 53 26@2x

### DIFF
--- a/components/Inspector.tsx
+++ b/components/Inspector.tsx
@@ -1,7 +1,9 @@
 import React, { HTMLProps } from 'react';
 import { queryModuleCache } from '../lib/ModuleCache.js';
+import { PARAM_VIEW_MODE, VIEW_MODE_CLOSED } from '../lib/constants.js';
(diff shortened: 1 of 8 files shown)


broofa:
@jimmywarting: thoughts?

PR #186, opened 2023-11-26
feat: add graph analysis feature(s)

broofa:
Rather than just listing all the modules, owners, and licenses in the graph pane (which is pretty noisy), this PR switches to a UI that emphasizes graph analysis to help users understand (and take action on?) dependency-related issues.  This adds "analyzers" for the following:

* Modules that appear more than once due to incompatible version constraints
* Modules that are deprecated
* Modules with only one owner
* Modules without a declared license (or that are explicitely "UNLICENSED")
* Modules with "discouraged" licenses
* Modules with "obsolete" licenses

I'm sure there's improvements that can be made, but this sets up a framework for doing this sort of analysis, which is something I've had in mind for awhile.

Play around with it at https://npmgraph-git-analysis-npmgraph.vercel.app/?q=jarvis

Thoughts, @fregante?

CleanShot 2023-11-26 at 15 34 01


guest271314:
I don't see where there is a `.json` export.

broofa:
> I don't see where there is a `.json` export.

That's not a planned feature.  I only pointed you here to note that the "All Modules" list enumerates each version of each module rather than grouping versions of the same module together the way the previous version does.

PR #193, opened 2023-12-20
Add URL parameter demo code

fregante:
This is wrong, how do you generate this URL?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ To graph a custom package.json module, provide the package contents here (in the
 
 https://npmgraph.js.org/?q=my_package%400.0.1#packages=%5B%7B%22name%22%3A%22my_package%22%2C%22version%22%3A%220.0.1%22%2C%22dependencies%22%3A%7B%22send%22%3A%220.18.0%22%7D%7D%5D
 
+```js
+'https://npmgraph.js.org/?q=my_package#' + new URLSearchParams([['packages', JSON.stringify(packageJson)]]).toString()
+```
+
 ### `select` (hash param)
 
 Select a module or category of modules.
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ To graph a custom package.json module, provide the package contents here (in the
 https://npmgraph.js.org/?q=my_package%400.0.1#packages=%5B%7B%22name%22%3A%22my_package%22%2C%22version%22%3A%220.0.1%22%2C%22dependencies%22%3A%7B%22send%22%3A%220.18.0%22%7D%7D%5D
 
 ```js
-'https://npmgraph.js.org/?q=my_package#' + new URLSearchParams([['packages', JSON.stringify(packageJson)]]).toString()
+'https://npmgraph.js.org/?q=my_package#' +
+  new URLSearchParams([['packages', JSON.stringify(packageJson)]]).toString();
 ```
 
 ### `select` (hash param)


README.md
```diff
@@ -53,6 +53,10 @@ To graph a custom package.json module, provide the package contents here (in the
 
 https://npmgraph.js.org/?q=my_package%400.0.1#packages=%5B%7B%22name%22%3A%22my_package%22%2C%22version%22%3A%220.0.1%22%2C%22dependencies%22%3A%7B%22send%22%3A%220.18.0%22%7D%7D%5D
 
+```js
+'https://npmgraph.js.org/?q=my_package#' + new URLSearchParams([['packages', JSON.stringify(packageJson)]]).toString()
```
fregante, line 57:
```suggestion
'https://npmgraph.js.org/?q=my_package#' + new URLSearchParams([['packages', JSON.stringify([packageJson])]]).toString()
```

fregante: A couple of notes for https://github.com/npmgraph/npmgraph/pull/157

- The package name should not be required
- The package name should not need to be repeated in `q` if it exists in the hash
- The package name should not appear again in the tree, it looks like noise, it already appears at the left https://npmgraph.js.org/?q=webext-tools

broofa: > * The package name should not be required
> * The package name should not need to be repeated in q if it exists in the hash

The package name is actually required given how the `q` and `packages` params work currently.  I wanted to support graphs  [like this one](https://npmgraph.js.org/?q=module-a%401.0.0#packages=%5B%7B%22name%22%3A%22module-a%22%2C%22version%22%3A%221.0.0%22%2C%22dependencies%22%3A%7B%22module-b%22%3A%221.0.0%22%2C%22module-c%22%3A%221.0.0%22%7D%7D%2C%7B%22name%22%3A%22module-b%22%2C%22version%22%3A%221.0.0%22%2C%22dependencies%22%3A%7B%22module-c%22%3A%221.0.0%22%2C%22uuid%22%3A%22*%22%7D%7D%2C%7B%22name%22%3A%22module-c%22%2C%22version%22%3A%221.0.0%22%2C%22dependencies%22%3A%7B%22module-a%22%3A%221.0.0%22%7D%7D%5D) ... where not only might the top-level module (`module-a`) be private (i.e. "uploaded"), but where there may be private modules elsewhere in the graph (`module-b`, `module-c`).

To make this work, the `q` and `packages` params need to be distinct from one another.  Specifically:
* `packages` - specifies package.json structs for all private modules that appear in the graph.
* `q` - specifies top-level modules (by package name)

Hence, why packages need to have a name... otherwise there's no way to reference them.

If the goal is to allow the upload of nameless packages (which I totally get why that would be desirable!), then we probably need a param that has a semantic of "top level entry point + its dependencies".  I'm open to suggestions for how that might work.

(Note: If we're going to support the nameless package case, it'd also be nice to be able to support other, less-structured use-cases like copy/pasting the contents of a `dependencies` structure.  'Have definitely had occasions where that would be helpful.)

> The package name should not appear again above the tree, it looks like noise

Yeah, it's a bit silly-looking when there's only a single module in the graph.  It probably makes sense to remove the title in that case.

But the title does help when there's [more than one entry point](https://npmgraph.js.org/?q=minimatch,cross-env,rimraf).

> You can actually repeat parameters multiple times...

That's true for _query_ parameters.  But `packages` (and all the other view-related params) are part of the URL`#hash`, which isn't actually part of the formal, named query params.  It's just an unstructured string value that, for simplicity, I parse as a JSON object.

broofa (approved):

README.md
broofa: > It's just an unstructured string value that, for simplicity, ~~I parse as a JSON object.~~

Yeesh... I'm an idiot.  The hash is structured using standard query param encoding, not JSON.   It's only the `packages` _value_ that's JSON-encoded. 🙄 

So... okay... it'd probably make sense to change that to the suggested  `package[]` notation.

Also, you mention this "kinda" blocks https://github.com/npmhub/npmhub/pull/179.  Is the issue there the `name` requirement, or something else?

fregante: > where not only might the top-level module (module-a) be private (i.e. "uploaded"), but where there may be private modules elsewhere in the graph (module-b, module-c).

I still don't see how that makes the name required? You can just pick the first package in the `packages` array in every case.

In any case, that's only a subset of usages, whereas the most common usages (single-package display/upload). If I upload a single nameless package then it's not required because it has no name, by definition.

> we probably need a param that has a semantic of "top level entry point + its dependencies". I'm open to suggestions for how that might work.


`dependencies[0]`

>  less-structured use-cases like copy/pasting the contents of a dependencies structure. 'Have definitely had occasions where that would be helpful.)

Sounds useful too, but that sounds like a job for the paste-box, not the URL parameter. The paste box receives some data and it adjusts it into a package.json-like structure to place it in the URL.

fregante: > But the title does help when there's [more than one entry point](https://npmgraph.js.org/?q=minimatch,cross-env,rimraf).

I see that already:

Screenshot

Even if one was a dependency of the other, it would not add a whole lot of information to the graph https://npmgraph.js.org/?q=webext-content-scripts%2C+webext-patterns

And here it doesn't do anything, even if I specifically request two packages: https://npmgraph.js.org/?q=webext-patterns%2C+escape-string-regexp

Screenshot 12

fregante: > Also, you mention this "kinda" blocks [npmhub/npmhub#179](https://github.com/npmhub/npmhub/pull/179)

The issue there was to figure out the correct format for the URL and also to actually support nameless packages.

broofa: This is all good input, thanks!   I'm gonna break this out into a few, more-focused, issues with some followup thoughts and we can go from there.

PR #201, opened 2023-12-31
Fix return value for hasExpectedFeatures

preethamrn:
Previously this method would always return false regardless of whether the expected features are found or not.

Fixes a regression in https://github.com/npmgraph/npmgraph/pull/200


broofa:
Good catch, thanks!

PR #225, opened 2024-07-13
Sidebar toggle animation fix

fregante:
See jumpy toggle (only visible when the graph is not empty, e.g. on https://npmgraph.js.org/?q=express)

https://github.com/user-attachments/assets/99b4e759-3624-4825-b974-84cc25727b3f

### DIFF
--- a/components/Inspector.scss
+++ b/components/Inspector.scss
@@ -1,6 +1,6 @@
 #inspector {
   display: flex;
-  flex-basis: 0;
+  flex: 0 0;
   flex-direction: column;
   transition: flex-basis 0.5s;
   height: 100vh;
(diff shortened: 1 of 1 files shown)


components/Inspector.scss
```diff
@@ -1,14 +1,14 @@
...
+  flex: 0 0;
   flex-direction: column;
   transition: flex-basis 0.5s;
   height: 100vh;
   overflow: hidden;
   color: var(--text);
 
   &.open {
-    flex: 0 0 var(--inspector-size);
+    flex-basis: var(--inspector-size);
```
fregante, line 11: This was overriding the whole of `flex`, but only `flex-basis` has `transition`

PR #224, opened 2024-07-13
Slightly rework sidebar style

fregante:
*screenshots coming*

PR #227, opened 2024-07-13
Keeps tabs visible while loading

fregante:
When the loader overlaps the tabs, it's impossible to do anything, you just have to wait. This is particularly a problem when the package is large: https://npmgraph.js.org/?q=web-ext

Screenshot 15

This PR just moves the bar to the bottom as a quick fix:

Screenshot

Combined with #224 it will look better down there.

### DIFF
--- a/components/App/Loader.scss
+++ b/components/App/Loader.scss
@@ -1,6 +1,7 @@
 .loader {
   position: absolute;
   right: 0;
+  bottom: 0;
   z-index: 99;
   box-sizing: border-box;
   background: var(--tabs-bg) var(--bg-gradient);


fregante:
I also considered moving it here but it's a bit harder to make that work since it would then overlap with the sidebar, etc

Screenshot 16

broofa:
Yeah, this has annoyed me at times in the past.   Never enough to do something about it, though.   So thanks for fixing this!

> I also considered moving it here but it's a bit harder to make that work since it would then overlap with the sidebar, etc

Sure.  Or it could go in the middle of the graph.   Or centered at the top of the graph... or .. or ... or.  🤷

PR #229, opened 2024-07-13
Fix first-load scroll on Safari and Firefox

fregante:
- Fixes https://github.com/npmgraph/npmgraph/issues/226

## Demo (Safari) ✅

Screenshot 20

### DIFF
--- a/components/GraphDiagram/GraphDiagram.tsx
+++ b/components/GraphDiagram/GraphDiagram.tsx
@@ -48,6 +48,23 @@ export type ZoomOption =
   | typeof ZOOM_FIT_WIDTH
   | typeof ZOOM_FIT_HEIGHT;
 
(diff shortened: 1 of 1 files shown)


broofa:
Something's not working right here (still).  If I select the `env-paths` modules [in this graph](https://npmgraph-git-scroll-on-safari-broofas-projects.vercel.app/?q=npm#select=exact%3Aenv-paths%402.2.1), it center the view on `env-paths` as expected:



... but if I then reload the page, it doesn't even come close to it:


This isn't exactly a regression since there's a similar issue on npmgraph.js.org currently.   But if we're trying to fix first-load scrolling... well... we should probably actually _fix_ it. :-/

broofa:
> The other issue you mentioned should be reported/fixed separately

Fair enough... damn it.  😉   Thanks for the fixes!

PR #230, opened 2024-07-13
chore: replace local DOM helper with native alternative

fregante:
@broofa I mistakenly pushed this proposal to `main`, but then immediately pushed a revert commit because I was unable to force-push it. Feel free to clean up the history on `main` before reviewing this.

See comments below to see what this PR is about.


fregante (commented):
This PR makes 3 changes:

- introduces a typescript selector parser, so you don't have to manually specify the selected element: https://github.com/g-plane/typed-query-selector
- splits "single element" queries from multi-element queries. This is done via $/$$, which is also used in the console: https://developer.chrome.com/docs/devtools/console/utilities#querySelectorAll-function
- drops the custom dom/elementSet helpers, which in many cases aren't much shorter or return the wrong type (e.g. `querySelector` does not guarantee the return, so it should always be `| null`)

components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -92,11 +92,11 @@ export default function GraphDiagram({ activity }: { activity: LoadActivity }) {
   async function handleGraphClick(event: React.MouseEvent) {
     const target = event.target as HTMLDivElement;
 
-    if ($('#graph-controls').contains(target)) return;
+    if ($('#graph-controls')!.contains(target)) return;
 
-    const el = $.up<SVGElement>(target, '.node');
+    const el = target.closest('g.node');
```
fregante, line 97: $.up was used 2 times. Arguably `A.closest(TYPE)` is more understandable

components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -92,11 +92,11 @@ export default function GraphDiagram({ activity }: { activity: LoadActivity }) {
...
 
-    const moduleKey = el ? $(el, 'title')?.textContent?.trim() : '';
+    const moduleKey = el ? el.querySelector('title')?.textContent?.trim() : '';
```
fregante, line 99: The new `$` helper is a simple bound `document.querySelector` call, so it can't be used in these cases. 

If you'd like, I can restore the previous `$(TYPE, PARENT)` signature. I actually have a similar well-typed module for this, but I haven't used it:

https://github.com/fregante/select-dom/blob/afdd637791737fc9dbcc8590d6655246d16bb4d5/index.ts#L13-L36

components/GraphDiagram/GraphDiagramDownloadButton.tsx
```diff
@@ -45,7 +45,7 @@ function downloadPng() {
     return;
   }
 
-  const canvas = $.create<HTMLCanvasElement>('canvas');
+  const canvas = document.createElement('canvas');
```
fregante, line 48: Another helper was `$.create`, which especially considering the required type assertion, was actually counter-productive

components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -383,18 +383,18 @@ async function colorizeGraph(svg: SVGSVGElement, colorize: string) {
 
   if (!colorizer) {
     // Unset all node colors
-    $(svg, 'g.node path').attr('style', undefined);
+    svg.querySelector('g.node path')!.removeAttribute('style');
```
fregante, line 386: Probably regression

components/GraphDiagram/GraphDiagram.tsx
broofa: Yeah, let's go with `select-dom` so we can have consistent use of $() and $(), throughout.

(Will that let us remove the `typed-query-selector` dependency if we're no longer calling querySelector/All directly?)

components/GraphDiagram/GraphDiagramDownloadButton.tsx
broofa: Agreed. npmgraph was my first real foray into Typescript.  APIs like `$.create()` weren't particularly nuanced as a result. 🤷

broofa (changes requested):

components/GraphDiagram/GraphDiagram.tsx
fregante: > remove the `typed-query-selector` dependency

You still want it for the .closest calls

components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -89,14 +89,14 @@ export default function GraphDiagram({ activity }: { activity: LoadActivity }) {
   // Signal for when Graph DOM changes
   const [domSignal, setDomSignal] = useState(0);
 
-  async function handleGraphClick(event: React.MouseEvent) {
-    const target = event.target as HTMLDivElement;
+  async function handleGraphClick(event: React.MouseEvent<HTMLElement>) {
+    const target = event.currentTarget;
```
fregante, line 93: I also remove these assertions and used native React types.

One thing to keep in mind is that `target` is the exact top-most element clicked, while `currentTarget` is the element where the listener was registered. From a quick look, it seems that this is fine.

components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -89,14 +89,14 @@ export default function GraphDiagram({ activity }: { activity: LoadActivity }) {
...
 
-    if ($('#graph-controls').contains(target)) return;
+    if ($('#graph-controls')!.contains(target)) return;
 
-    const el = $.up<SVGElement>(target, '.node');
+    const el = target.closest('g.node');
 
-    const moduleKey = el ? $(el, 'title')?.textContent?.trim() : '';
+    const moduleKey = el ? $('title', el)?.textContent?.trim() : '';
```
fregante, line 99: Note that select-dom, like the console API, takes the parent as the second parameter.

components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -383,18 +383,20 @@ async function colorizeGraph(svg: SVGSVGElement, colorize: string) {
 
   if (!colorizer) {
     // Unset all node colors
-    $(svg, 'g.node path').attr('style', undefined);
+    for (const node of svg.querySelectorAll('g.node path')) {
+      node.removeAttribute('style');
+    }
```
fregante, line 388: One thing to be aware of is that the previous code silently supported multiple elements, while $$ does not. That is `$('a').remove()` would remove all `a` elements, but that requires an explicit loop now.

I think this was the only instance that needed it.

broofa:
Hmm.  Click-to-select modules in the graph appears to be broken.

CleanShot 2024-07-19 at 06 42 36

fregante:
Fixed. One of the `event.target` changes was significant since it used event delegation

x

components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -367,11 +365,8 @@ export function updateSelection(
   }
 
   // Set selection classes for edge elements
-  for (const titleEl of [...$<SVGElement>('svg .edge')]) {
-    const edgeTitle = $(titleEl, '.edge title')?.textContent ?? '';
-    const edge = $.up<SVGPathElement>(titleEl, '.edge');
-    if (!edge) continue;
-
+  for (const edge of $$('svg g.edge')) {
+    const edgeTitle = $('title', edge)?.textContent ?? '';
```
fregante, line 369: Also fixed a bug here, it was not a `path` element so my `path.edge` selector was failing.


<table>
<tr>
	<th>Regression
	<th>Fixed
<tr>
	<td>Screenshot
	<td>Screenshot 1
</table>







I'm not sure why it was selecting `.edge`, then `.edge title`, then using `$.up` to find `.edge`. I dropped that part and just used the first `.edge`

PR #237, opened 2024-07-19
Add type-checking

fregante:
I noticed that types weren't actually being verified, but were only visible in the IDE and to ESLint


fregante:
uh again I merge it without adding `chore:` 🥲

broofa:
> uh again I merge it without adding `chore:` 🥲

I need to add a pre-commit check for this... some day.

fregante:
CI check added in #259. pre-commits don't prevent opening PRs with the wrong title

PR #238, opened 2024-07-21
feat: revisit color scheme

fregante:


components/App/Loader.scss
```diff
@@ -34,8 +34,8 @@
       -45deg,
       white 0,
       white 20px,
-      var(--highlight) 20px,
-      var(--highlight) 40px
+      var(--accent) 20px,
+      var(--accent) 40px
```
fregante, line 38: Matching terminology: https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color

components/GraphDiagram/GraphDiagram.scss
```diff
@@ -95,20 +96,24 @@ g.node {
...
+    }
+    & > text {
+      fill: var(--text-on-accent);
     }
   }
 
   &.upstream {
     & > path {
-      stroke: var(--stroke-blue);
+      stroke: var(--accent-100);
```
fregante, line 110: Dependents: Same as accent, but lighter:

components/GraphDiagram/GraphDiagram.scss
```diff
@@ -95,20 +96,24 @@ g.node {
...
     }
   }
 
   &.downstream {
     & > path {
-      stroke: var(--stroke-violet);
+      stroke: var(--accent);
```
fregante, line 116: Dependencies: they're still part of the selected module, so I believe they should have the same color. 

The selected module is now fully orange so it's still visually different.

components/Inspector.scss
```diff
@@ -2,8 +2,9 @@
   display: flex;
   flex: 0 0;
   flex-direction: column;
-  transition: flex-basis 0.5s;
-  box-shadow: 0px 7px 3px var(--inset-shadow-color, rgba(214, 214, 214, 0.467));
+  transition: flex-basis var(--transition-duration);
+  border-top-left-radius: var(--rad_lg);
+  border-bottom-left-radius: var(--rad_lg);
```
fregante, line 7: I think https://github.com/npmgraph/npmgraph/pull/224 made it a little too _flat_ so restoring some rounded corners and moving the shadow back inside livened it up a bit:

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 12
	<td>Screenshot 11
</table>

components/Inspector.scss
```diff
@@ -52,39 +53,27 @@ label {
...
     line-height: var(--tabs-height);
   }
-
-  .tab {
-    cursor: pointer;
-    border-left: solid 1px #888;
-    background: var(--tabs-bg) var(--bg-gradient);
-    padding: 0 10px;
-    font-weight: 600;
-    font-size: 12pt;
```
fregante, line 68: Redundant styles

components/Splitter.scss
```diff
@@ -3,10 +3,13 @@
   top: 0;
   right: 0;
   z-index: 1;
-  cursor: pointer;
+  transition: border-radius var(--transition-duration);
+  cursor: default;
```
fregante, line 7: Parts of the UI use a pointer, parts use the regular cursor. I think we should standardize it and use the default, the buttons already have a hover state so they don't need `pointer`:

https://github.com/tailwindlabs/tailwindcss/issues/8961#issuecomment-1533249212

components/Splitter.tsx
```diff
@@ -11,7 +14,7 @@ export function Splitter({
 }) {
   return (
     <div id="splitter" className="bright-hover tab" onClick={onClick}>
-      {isOpen ? '✗' : '\u{25c0}'}
+      {isOpen ? blackRightPointingTriangle : blackLeftPointingTriangle}
```
fregante, line 17: I changed my mind on the x, I think a chevron or triangle works best here.

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 13<td>Screenshot 14
</table>

components/Tab.scss
```diff
@@ -1,17 +1,20 @@
...
+  button {
+    all: unset;
+    display: flex;
+    align-items: center;
+    background: var(--tabs-bg) var(--bg-gradient);
+    padding: 0 10px;
+  }
+
+  &.active button {
+    background: var(--accent);
```
fregante, line 17: I reworked the DOM to fix the borders. The contrast was poor and one went missing on `hover` and `active`

Screenshot 15

Screenshot 16

index.scss
```diff
@@ -23,7 +25,13 @@
 
   --tabs-bg: #7e7e7d;
   --tabs-text: #fff;
-  --highlight: #e65c00;
+  --tabs-border: rgba(203, 203, 203, 0.796);
+  --accent: #da6200;
```
fregante, line 29: The accent color was slightly darkened to improve contrast of "white on orange"

components/GraphDiagram/GraphDiagram.scss
broofa: Can we make the distinction between upstream vs. downstream dependencies more obvious?  It's pretty subtle at the moment and I've seen graphs with circular dependencies where being able to tell the difference is really helpful.  (Sorry, can't seem to find a good example at the moment.)

Seeing that difference was the initial motivation for the blue-violet colors.  Although, admittedly, I chose those colors while developing the feature just to make them *blatantly* obvious, then never got around to finessing them so... yeah... I totally understand your reaction.

Maybe use slightly different hues of orange?   E.g. make `upstream` be `color-mix(in hsl shorter hue, var(--accent) 25%, var(--stroke-yellow))`?

components/GraphDiagram/GraphDiagram.scss
```diff
@@ -52,7 +52,8 @@
     line-height: 1em;
 
     &.selected {
-      background-color: var(--highlight);
+      background-color: var(--accent);
+      color: var(--text-on-accent);
```
broofa, line 56: I like the highlight->accent name change. 💯 

Did you intend to lose the `warning` style for deprecated modules when selected...
before:
CleanShot 2024-07-21 at 10 16 41

after:
CleanShot 2024-07-21 at 10 18 24

Also, the white-on-orange selection text is a little harder to read.  Not a lot, but it's enough that I notice the difference with my eyesight.  I know we have this style elsewhere in the UI (tabs, zoom buttons), but in those places the text is larger and not as dynamic.

components/GraphDiagram/GraphDiagram.scss
broofa: Yeah, I agree having the dependency colors closer to, if not the same, as the selection color makes sense.

components/Splitter.tsx
broofa: No strong opinion on this.

broofa (commented):
Commenting as opposed to "requesting changes" because I'm not entirely sure there's much to be done about my feedback.  Curious as to your reaction. :-)

components/GraphDiagram/GraphDiagram.scss
fregante: As a reaction to huge dependency trees caused by a certain user, people have been posting npmgraph screenshots of the before/after tree cleanup. It'd be nice to have those trees look good, that's why I wanted to remove the cyan/magenta colors.

I'll run some tests and see how to generate an appealing palette from the accent.

components/GraphDiagram/GraphDiagram.scss
fregante: Dejavu, pretty sure I made the same exact change a few years ago. I don’t think it's a huge issue honestly, because the selected item is shown in the sidebar, which is even more explicit in telling you what that color means:

Screenshot 7

components/GraphDiagram/GraphDiagram.scss
fregante: By the way, do you have an example of a tree with circular dependencies? It would be good to have a list of such dependencies to test out certain features. They could be added to the contributing.md file.

broofa: Alas, not at the moment.  I'll update here the next time I see one.  I typically see them in projects (as opposed to published modules) where plugins and dependencies tend to be a bit messier.

broofa: Here's a [somewhat contrived] example: https://npmgraph.js.org/?q=string-width%404.2.3%2C+yargs%4016%2C+wrap-ansi%407%2C+#select=exact%3Acliui%407.0.4

Notice how upstream/downstream nodes and edges kind of blend together here ...

CleanShot 2024-07-22 at 09 59 52

... versus what's currently on `main` (which is admittedly a bit over-the-top colorwise :-) )

CleanShot 2024-07-22 at 10 00 36

fregante:
I used https://mycolor.space/?hex=%23DA6200&sub=1 to create 2 similar colors:



Pulling the cyclic dependency link out: https://npmgraph.js.org/?q=string-width%404.2.3%2C+yargs%4016%2C+wrap-ansi%407%2C+#select=exact%3Acliui%407.0.4

<table>
<tr>
	<td>Screenshot 5
	<td>Screenshot 4
<tr>
	<td>Screenshot 2
	<td>Screenshot 3
</table>

I'm not totally sure about the green, but it's slightly more pleasant. 

I also reverted the orange background, even though I think it's a downgrade. I'll keep playing with the colors.

fregante:
Ok I think I got it:

- selected: orange, thick
- downstream: orange
- upstream: cyan (but the direct negative of the accent)

<table>
<tr>
	<td>Screenshot 7
	<td>Screenshot 8
</table>

fregante:
Seems obvious in retrospect: unrelated modules are already dimmed, so why does the upstream need to use a third color?

Screenshot

broofa (approved):

fregante:
Merging now. If you'd like I can drop unused colors separately.

To create related colors we can use `oklch` colors and `color-mix()` in the future:

- https://oklch.com/

PR #242, opened 2024-07-21
chore: use Biome instead of eslint and prettier

fregante:
Just trying it out since eslint decided to cause trouble (flat config, eslint v9 being breaking). I ran the migration scripts and they were pretty straightforward.

Feel free to close if you prefer eslint/prettier.

Cons:

- yml, md, and "css order" are not yet supported. I could leave Prettier to address these only
- not all lint rules are included
- lint config is currently a bit long, auto-generated. It's possible that `recommended: true` is actually enough
 
Good news:

- there were zero regressions on the current code


broofa:
Hmm... solution in search of a problem?  Biome's main selling point seems to be performance, which isn't a concern here (right?)

The verbose config file is a bit of a code smell for me.  I've found complex config files are harder to maintain long-term.  What does this PR look like if you go with as barebones a config file as possible?

(FWIW, my inclination at the moment is to just close this.  So I don't know that I'd put a lot of effort into it.  But I would be curious to see what Biome does to the codebase with a minimal config setup.)

fregante:
> search of a problem

I noticed that you recently updated all dependencies… except ESLint, why? 

This PR is the answer to that problem

fregante:
Ok so I cleanup the config a bit and now all that's left in it are a few _new_ useful rules. A couple of them slipped by and I applied their auto-fixes.

I think fixing the remaining rules would actually be a good reason to upgrade. Biome comes with many rules that are normally part of other plugins, like the typescript, unicorn, a11y, react plugins. Overall, even judging by the unified JS/CSS configuration, Biome has a better architecture.

fregante:
> I would be curious to see what Biome does to the codebase with a minimal config setup

See https://github.com/npmgraph/npmgraph/pull/243

I'd be open to fix the remaining issues or just close both PRs. I think it's an improvement overall, but it's your project so it's up to you.

fregante:
If you accept this change, don't merge it just yet. Marking as draft for now.

Issue #222, opened 2024-07-13
Add search/filter in the graph


fregante:
The Graph tab has the ability to list and scroll to specific modules. It would be great to filter the "all modules" list and/or highlight specific items on the graph.

Currently this is possible via the browser's own "search on page" command, so it's not fundamental.

fregante:
For reference, I had asked this before, but I think that now with collapsible panels in the Graph/Report tab it would still be useful.

- https://github.com/npmgraph/npmgraph/issues/35

PR #244, opened 2024-07-21
Reorganize Graph/Info tabs

fregante:
The content and organization of the tabs could be improved I think. 

This is a proposal. 

I haven't renamed any files to match the contents and I haven't made code changes other than minor CSS cleanup around spacing and borders


fregante:
# Start

I don’t think there needs to be a dedicated "About" pane _plus_ the Start pane.


This change brings out some information to the first, cleaning up the style as well.

I feel that the page is missing some "npmgraph" branding but I haven't thought of what that could look like yet. Maybe a logotype instead of "Start". _Maybe another time…_

To do:

- [ ] Fix field size

Screenshot 11

fregante:
# Report

I like that the "Graph" tab has become its own exploratory tool unrelated to the graph. For this reason I think it needs its own dedicated tab. Its connection to the graph is coincidental.

For this reason I moved the view options to their own tab.

This tab would actually be a good place to introduce a search/filter field:

- #222 

Screenshot 10

fregante:
# Module

This tab is unchanged. I'm not a fan of it being empty, but let's leave it alone in this PR.

fregante:
# View

Nearly empty, but I think these options don't belong anywhere else. More things could come in the future though:

- the list of uploaded package.json? I don't remember if you still show them
- if renamed to "Options": https://github.com/npmgraph/npmgraph/issues/19


Screenshot 12

broofa:
Just a quick note to let you know I won't be able to get to this or other PRs until after next week.

fregante:
Style changes were extracted to:

- https://github.com/npmgraph/npmgraph/pull/271

Tabs can be addressed individually, I started with:

- https://github.com/npmgraph/npmgraph/pull/291#pullrequestreview-2233613006

PR #239, opened 2024-07-21
feat: revisit Module tab style

fregante:
Improvements:

- reduce space below package name
- move `QueryLink` to explicit link (it was never clear what clicking the package did)
- add icon to npm link
- add icon to package.json link
- reduce `ExternalLink` icons across the app (they were too big)
- always show the repo link, even if homepage is present (I never care about the `homepage`, I want the repo)

The plan was actually to move the npm/github/homepage icons next to the title, but on long package names they would wrap 

## Before

Screenshot 21

## After

Screenshot 20

PR #245, opened 2024-07-22
feat: allow opening modules in a new tab

fregante:
Sometimes I just want to explore a module in a new tab. This enables it via the standard cmd+click.

I think `metaKey` doesn't work on Windows so if accepted I'll have to do ua-sniffing and enable it on `ctrlKey` as well.

### DIFF
--- a/components/GraphDiagram/graph_util.ts
+++ b/components/GraphDiagram/graph_util.ts
@@ -218,7 +218,9 @@ export function composeDOT({
       fontsize *= Math.max(1, Math.log10(unpackedSize) - 2);
     }
 
-    const vs = { root: level == 0, fontsize };
+    const link = new URL(location.origin);
+    link.searchParams.append('q', module.key);
+    const vs = { root: level == 0, fontsize, href: link.href };
 
     nodes.push(`"${dotEscape(module.key)}" ${vizStyle(vs)}`);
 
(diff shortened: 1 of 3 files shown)


components/GraphDiagram/graph_util.ts
```diff
@@ -218,7 +218,9 @@ export function composeDOT({
       fontsize *= Math.max(1, Math.log10(unpackedSize) - 2);
     }
 
-    const vs = { root: level == 0, fontsize };
+    const link = new URL(location.origin);
+    link.searchParams.append('q', module.key);
+    const vs = { root: level == 0, fontsize, href: link.href };
```
fregante, line 223: This adds links to the graph itself instead of having to parse the clicked module from the graph and then using `window.open()`

fregante:
Also:

- [ ] turn "View" button into a link as well


The one added in #

broofa:
* [ ] This breaks simple click-to-select of modules:
CleanShot 2024-07-22 at 09 10 40

Also, from #239
* [ ] "View" button should probably not be enabled when the current root module is selected (user is already at that view)
* [ ] "View" button is not properly aligned
CleanShot 2024-07-22 at 09 15 13

fregante:
Should be good now. I can look into the View button separately, it's unrelated

x

broofa (approved):

components/GraphDiagram/graph_util.ts
fregante: It's `href` https://graphviz.org/docs/attrs/href/

PR #248, opened 2024-07-23
nit: rename "View" to "Go"

broofa:
@fregante 

"Speak now or forever hold your peace"... (waits half a second)... "okay, then." 😉


fregante:
Sounds good, and thanks for keeping the arrow to the left because I think that's what explained it best

PR #251, opened 2024-07-27
feat: stay on current tab unless the user requests a change

fregante:
I always find it jarring to have the sidebar change while I'm checking out multiple modules or comparing them:

https://github.com/user-attachments/assets/fc46cc48-4633-4319-8c81-1be00529fd09

Likewise, I don't think that the tab should switch to Graph when the user clicks on an empty area of the graph


fregante:
- https://github.com/npmgraph/npmgraph/issues/250

Together this means that clicking "Go" will now stay on the same tab:

- the selected module becomes the queried module
- the this pane would then point to the queried module
- no change happens here (other than the "Go" button disappearing)

fregante:
I can't stand the current behavior anymore 🥹 let's try this out and tweak it later if necessary

PR #252, opened 2024-07-27
fix: hide empty sections in Graph tab

fregante:
- Fixes https://github.com/npmgraph/npmgraph/issues/247

Screenshot 6

### DIFF
--- a/components/GraphPane/GraphPane.scss
+++ b/components/GraphPane/GraphPane.scss
@@ -17,3 +17,11 @@
 #colorize-key {
   margin-top: 1rem;
 }
+
+#report-section {
+  // Hide empty sections
+  h3:has(+ h3),
+  h3:last-child {
+    display: none;
+  }
+}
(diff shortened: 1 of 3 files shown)


components/GraphPane/reports/reporters/modulesAll.tsx
```diff
@@ -9,6 +9,10 @@ export function modulesAll({
   moduleInfos,
   entryModules,
 }: ModuleAnalysisState): RenderedAnalysis {
+  if (moduleInfos.size === 0) {
+    return;
+  }
```
fregante, line 14: This hides the "0 modules" section. Other reporters have similar code

components/GraphPane/GraphPane.scss
```diff
@@ -17,3 +17,11 @@
...
   margin-top: 1rem;
 }
+
+#report-section {
+  // Hide empty sections
+  h3:has(+ h3),
+  h3:last-child {
+    display: none;
+  }
+}
```
fregante, line 27: I think it's not easy to detect whether a section is empty in React, so I thought I'd just wrap them in a div and use CSS.

fregante:
Merging this because it's a quick fix. It would have to be rethought after this is merged because this "Report" tab would then be completely empty.

PR #253, opened 2024-07-27
fix: hide Go button if there's nothing to change

fregante:
- Closes https://github.com/npmgraph/npmgraph/issues/246

## Normal (non-entry module selected)

Screenshot 8

## No Go (entry module selected)

Screenshot 7

## Normal (entry module selected, but not the only entry module)

Screenshot 9

### DIFF
--- a/index.scss
+++ b/index.scss
@@ -28,7 +28,7 @@
   --tabs-text: #fff;
   --tabs-border: rgba(203, 203, 203, 0.796);
 
-  --accent: #e65c00;
+  --accent: #da6200;
   --text-on-accent: #fff;
 
   // Shadow colors depend on the background color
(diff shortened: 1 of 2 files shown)


index.scss
```diff
@@ -28,7 +28,7 @@
   --tabs-text: #fff;
   --tabs-border: rgba(203, 203, 203, 0.796);
 
-  --accent: #e65c00;
+  --accent: #da6200;
```
fregante, line 31: Also slipping this change back in because I regret reverting it in a previous PR:

[`971e1f6` (#238)](https://github.com/npmgraph/npmgraph/pull/238/commits/971e1f6e3db459504e62bb8992d6a4d1312aa71c)

broofa (approved):

Issue #240, opened 2024-07-21
Accept package.json drops all over the page


fregante:
1. Pick a file
2. Drag it over the page
3. Miss the target by a few pixels
4. The app is gone, the json file is shown in the browser

Pages that accept files on drop generally accept them on the whole page, or at least block drops that were dropped on the wrong target.

I'd move these lines to the main App myself, but I don't know how to connect them to the `FileUploadControl`  myself:

https://github.com/npmgraph/npmgraph/blob/867ea561c440de204048b5c14df73aff66ead13a/components/InfoPane/FileUploadControl.tsx#L112-L114

PR #258, opened 2024-07-28
feat: paste JSON text and JSON files 

fregante:
- Closes https://github.com/npmgraph/npmgraph/issues/97

Behaviors, these only apply to the Start tab:

- valid JSON strings and files are accepted anywhere on the page and don't appear in the search field, and a "flash" notification is shown
- invalid JSON strings and files produce a notification only if the search field is not focused. This allows the user to paste whatever in the search field as expected

I made this part of the FileUploadControl because most of the logic was already in there, but I'd expect all of this code to be moved out for:

- #240


components/InfoPane/InfoPane.tsx
```diff
@@ -7,24 +7,26 @@ import QueryInput from './QueryInput.js';
...
+        <li>
+            Multiple, versioned module names: <QueryLink query={['cross-env@6', 'rimraf']}/>
+        </li>
+        <li>
+          A URL to a <QueryLink query="https://github.com/npmgraph/npmgraph/blob/main/package.json">
+            package.json file
+          </QueryLink>
+        </li>
+        <li>package.json pastes (copied as text or file)</li>
+      </ul>
```
fregante, line 29: New look to allow a compact explanation of the multiple input modes

Screenshot

The drop area will also be turned into:

- package.json drag and drop anywhere on the page
- package.json [picked](#select) from your file system


after:

- #240

fregante, line 29: New look to allow a compact explanation of the multiple input modes

Screenshot

The drop area will also be turned into:

> - package.json drag and drop anywhere on the page
> - package.json [picked](#select) from your file system


after:

- #240

fregante: Documentation changes extracted to their own PR:

- https://github.com/npmgraph/npmgraph/pull/279

components/InfoPane/InfoPane.tsx
```diff
@@ -27,6 +27,7 @@ export default function InfoPane(props: HTMLProps<HTMLDivElement>) {
             package.json file
           </QueryLink>
         </li>
+        <li>package.json pastes (copied as text or file)</li>
       </ul>
 
       <FileUploadControl />
```
fregante, line 33: And then, after:

- #240


The drop area will also be turned into two bullet points on the list above:

> - package.json drag and drop anywhere on the page
> - package.json [picked](#select) from your file system

components/InfoPane/FileUploadControl.tsx
```diff
@@ -52,23 +61,18 @@ export default function FileUploadControl(props: HTMLProps<HTMLLabelElement>) {
...
     try {
-      pkg = JSON.parse(content);
+      pkg = JSON.parse(json);
     } catch {
-      flash(`${file.name} is not a valid JSON file`);
+      flash(`${filename ?? 'Pasted content'} is not a valid JSON file`);
       return;
     }
 
+    flash(`Loading ${filename ?? 'pasted content'}...`);
```
fregante: I dropped this line. I was looking for a way to give confirmation that the tool is _doing something_ but it's sometimes confusing, because this notification lasts longer than it takes to actually parse the file (which is 0ms in case of repeated pastes)

We might have to think of better ways to give feedback, for example by implementing mandatory delays—that is, _always_ show a minimum "loading" state so that the user knows something is happening, like via https://github.com/smeijer/spin-delay

PR #262, opened 2024-07-28
chore: bump eslint to v9

fregante:
I tried to keep the diff at a minimum. Most are simple autofixes. In some cases I also used `document.title` and `.textContent` (instead of `.innerText`) manually because no autofix was available.

I also used a greater shared config that includes multiple plugins as dependencies, it's much easier to handle them this way. I just picked a decent maintained config. I'm partial to XO but it's late on the flat config train.

https://github.com/antfu/eslint-config

The config also includes React-related rules but they're not enabled in this PR.


eslint.config.mjs
```diff
@@ -0,0 +1,25 @@
...
+    'style/jsx-one-expression-per-line': 'off',
+    'style/member-delimiter-style': 'off',
+    'style/multiline-ternary': 'off',
+    'style/operator-linebreak': 'off',
+    'style/quote-props': 'off',
+    'style/quotes': 'off',
+    'style/semi': 'off',
+    'style/type-generic-spacing': 'off',
+    'ts/consistent-type-imports': 'off',
+    'ts/consistent-type-definitions': ['error', 'type'],
```
fregante, line 23: Most of these were conflicting with Prettier or created large diffs. I can apply them later.

package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "npmgraph",
-  "version": "3.1.1",
   "type": "module",
+  "version": "3.1.1",
+  "description": "Visualize npm dependency graphs (public and private!)",
```
fregante, line 5: The linter also sorts package.json now

tsconfig.json
```diff
@@ -1,12 +1,12 @@
 {
   "compilerOptions": {
+    "target": "ES2023",
     "jsx": "react-jsx",
-    "lib": ["ES2022", "DOM", "DOM.Iterable"],
+    "lib": ["ES2023", "DOM", "DOM.Iterable"],
```
fregante: TS 5.4 should have Map.groupBy but I don't know how to include its type. 🤷‍♂️

fregante:
Merging as it's just some lint config change

PR #264, opened 2024-07-31
chore: cleanup lint config and apply new rules

fregante:
- Continues #262 

This PR:

- Enables import rules and applies their autofixes (`type` keyword, import order)
- Organizes config to highlight prettier incompatibilities
- _Manually_ applies `eqeqeq` fixes (about 15 replacements)

I also wanted to include new React rules but the PR has a bunch of autofixes already, those would get lost.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,32 +1,38 @@
 import antfu from '@antfu/eslint-config';
 
-export default antfu({
-  rules: {
-    eqeqeq: 'off',
-    'no-alert': 'off',
-    'no-console': 'off',
-    'antfu/if-newline': 'off',
-    'import/order': 'off',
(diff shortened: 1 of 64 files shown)


eslint.config.mjs
```diff
@@ -19,7 +20,14 @@ export default antfu({
...
-    'ts/consistent-type-definitions': ['error', 'type'],
+  },
+  typescript: {
+    overrides: {
+      // https://github.com/antfu/eslint-config/issues/570
+      'ts/consistent-type-imports': [
+        'error',
+        { fixStyle: 'inline-type-imports' },
+      ],
+    },
```
fregante, line 31: This override is separate at least until:

- https://github.com/antfu/eslint-config/issues/570

PR #266, opened 2024-07-31
feat: improve mobile support

fregante:
Main changes:

- re-aligns the zoom buttons
- avoids `position:fixed` on the whole #app, because it clips the view at the bottom on mobile
- display vertically on portrait screens. Not perfect, but so much better!
- ~~restyles the zoom buttons~~ _Edit: Dropped from this PR for now, ignore the screenshot_


fregante (commented):
- [x] Fix unwanted change on desktop:

scroll

fregante:
A few improvements:

- both the graph and the inspector have a minimum height, this helps on small graphs an empty sections ("Module" tab), so the UI doesn't jump around so much
- the inspector is always visible on mobile, even if accessed via `#hide=` URL
- the loader is now `display: fixed`

I think if we want to improve the situation further we should place the graph into a tab of the inspector and keep the tabs always visible on top, but that'd require a lot of changes and a rethinking of what happens when you select a module on the graph. Not worth it

fregante:
> having the inspector "below the fold" (out of sight) isn't great UI

If the core feature of npmgraph is the graph, then I don’t think it's a huge problem. It would benefit from a header however. Maybe after the logo we can add a logo + "show inspector" button in the header, just on mobile.




> otherwise it's difficult to discover.

If you land here, you _will_ scroll:

Screenshot 1

I considered keeping the tab bar visible but it greatly reduces the usability. I think a header would be more beneficial.

PR #271, opened 2024-08-05
feat: improve spacing and separation of sections

fregante:
- Extracted some style from #244 
- Improved general style, spacing and consistency of sections

This PR only changes some style, it doesn't move content. #244 is still open as a general organization suggestion (don't look at its code)

PR #273, opened 2024-08-05
fix: hide empty sections in Graph tab

fregante:
Reapplies 

- https://github.com/npmgraph/npmgraph/pull/252

After

- #271

### DIFF
--- a/components/GraphPane/GraphPane.scss
+++ b/components/GraphPane/GraphPane.scss
@@ -16,10 +16,8 @@
   margin-top: 1rem;
 }
 
-#report-section {
-  // Hide empty sections
-  h3:has(+ h3),
-  h3:last-child {
-    display: none;
-  }
+// Hide empty sections
+.report-section hr:nth-last-child(2),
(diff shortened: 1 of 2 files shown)


components/GraphPane/GraphPane.scss
```diff
@@ -16,10 +16,8 @@
...
-#report-section {
-  // Hide empty sections
-  h3:has(+ h3),
-  h3:last-child {
-    display: none;
-  }
+// Hide empty sections
+.report-section hr:nth-last-child(2),
+.report-section h3:nth-last-child(1) {
+  display: none;
```
fregante, line 22: It seems that there's no _React way_ to do this — which is kind of expected: children are rendered only if they're specified by the parent, but if the parent returns `null` then the children will not attempt to render. ♻️

fregante (commented):
<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 10
	<td>
Screenshot 11

</table>

PR #287, opened 2024-08-10
feat: add og:image

fregante:
- Mentioned in https://github.com/npmgraph/npmgraph/pull/285

It would be nice to use https://vercel.com/docs/functions/og-image-generation to create a custom image depending on the input, but it's low priority. I'd be more interested in changing the `title` depending on the `q` instead

Screen Shot 2024-08-10 at 12 30 29


fregante:
Demo:

Screenshot 2

PR #303, opened 2024-09-14
fix: add globalThis to unsupported feature checks

broofa:


fregante:
What is this for? Who encountered this issue?

lib/index.tsx
```diff
@@ -12,16 +12,28 @@ import { syncPackagesHash } from './ModuleCache.js';
...
+  }
+}
+
 // Various features we depend on that have triggered bugsnag errors in the past
 function detectFeatures() {
   const unsupported = new Map<string, JSX.Element>();
 
   // API checks
   const features = {
+    '||= (Logical Or)': isValidJS('a ||= 123'),
```
fregante, line 31: If you care about these users we can just increase the browserlist supported browsers. But this and `globalThis` have been available for several years (4 and 5 respectively) so I don’t think anyone encounters them.

If you’re doing this because of bugsnag I’d advise detecting bots and/or very old browser versions and disabling their error reporting instead.

PR #310, opened 2024-10-08
chore: add metatag for Google Search Console

fregante:
@broofa mind if I take a look a the stats for the website? Just curious how it's looking and if there are any warnings. You should probably also add your own verification


broofa:
> You should probably also add your own verification

Looks like I can use my existing gtag to verify.  'Just needed to move it to the HEAD to make Google happy.

PR #316, opened 2025-02-24
fix: correct downstream unpacked size calculation

bluwy:
TLDR: Shared dependencies were calculated twice for `Unpacked Size (module + dependencies):`. This PR fixes it to count once.

## Issue description

I was checking https://npmgraph.js.org/?q=magicast and I noticed that the `Unpacked Size (module + dependencies):` for `magicast` (root) isn't exactly accurate if you manually check the size of its dependencies.

It has a unique shared `@babel/types` dependency like this:
image

Before,`@babel/types` was accounted twice. Incorrectly totalling 8MB for `magicast`. This PR fixes to only account once with the new correct total being 5MB.

### DIFF
--- a/components/GraphDiagram/graph_util.ts
+++ b/components/GraphDiagram/graph_util.ts
@@ -142,9 +142,10 @@ export async function getGraphForQuery(
         const downstreamModule = await getModule(getModuleKey(name, version));
 
         const moduleInfo = await _visit(downstreamModule, level + 1);
+        if (!moduleInfo) return;
 
-        moduleInfo?.upstream.add({ module, type });
-        info?.downstream.add({ module: downstreamModule, type });
+        moduleInfo.upstream.add({ module, type });
+        info.downstream.add({ module: moduleInfo.module, type });
       }),
     );
 
@@ -278,11 +279,12 @@ export function foreachUpstream(
   callback: (module: Module) => void,
   seen: Set<Module> = new Set(),
 ) {
-  const info = graph.moduleInfos.get(module.key);
-  if (!info || seen.has(module)) return;
   seen.add(module);
+  const info = graph.moduleInfos.get(module.key);
+  if (!info) return;
 
   for (const { module } of info.upstream) {
+    if (seen.has(module)) continue;
     callback(module);
     foreachUpstream(module, graph, callback, seen);
   }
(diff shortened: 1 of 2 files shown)


components/GraphDiagram/graph_util.ts
```diff
@@ -142,9 +142,10 @@ export async function getGraphForQuery(
         const downstreamModule = await getModule(getModuleKey(name, version));
 
         const moduleInfo = await _visit(downstreamModule, level + 1);
+        if (!moduleInfo) return;
 
-        moduleInfo?.upstream.add({ module, type });
-        info?.downstream.add({ module: downstreamModule, type });
+        moduleInfo.upstream.add({ module, type });
+        info.downstream.add({ module: moduleInfo.module, type });
```
bluwy, line 148: The main change here is `module: moduleInfo.module` so that if `moduleInfo` was [already cached](https://github.com/npmgraph/npmgraph/blob/ceb40b728a064654842d43ec92641c9dec24a8bf/components/GraphDiagram/graph_util.ts#L118-L123), we reuse its module instance. Otherwise there is multiple instance of `Module` in the graph, which makes the `seen: Set<Module>` cache below not working correctly.

components/GraphDiagram/graph_util.ts
```diff
@@ -294,11 +296,12 @@ export function foreachDownstream(
...
   seen: Set<Module> = new Set(),
 ) {
-  const info = graph.moduleInfos.get(module.key);
-  if (!info || seen.has(module)) return;
   seen.add(module);
+  const info = graph.moduleInfos.get(module.key);
+  if (!info) return;
 
   for (const { module } of info.downstream) {
+    if (seen.has(module)) continue;
```
bluwy, line 304: The cache is handled differently here so that we only call the `callback` if the module was not already seen. Before it was calling the `callback` despite already seen.

components/GraphDiagram/graph_util.ts
```diff
@@ -278,11 +279,12 @@ export function foreachUpstream(
...
   seen: Set<Module> = new Set(),
 ) {
-  const info = graph.moduleInfos.get(module.key);
-  if (!info || seen.has(module)) return;
   seen.add(module);
+  const info = graph.moduleInfos.get(module.key);
+  if (!info) return;
 
   for (const { module } of info.upstream) {
+    if (seen.has(module)) continue;
```
bluwy, line 287: This change is in `foreachUpstream` to match the changes in `foreachDownstream` (which affects the calculation). This function doesn't seem to be used, but I updated to make it consistent.

components/GraphDiagram/graph_util.ts
broofa: This was a workaround for a bug that was allowing a module to have more than one entry in the cache.  I've just added the fix for that (`newCachedModule()`), and reverted this.

components/GraphDiagram/graph_util.ts
broofa: Yup, this method isn't used.  I probably added it thinking I'd have a use for it.  I don't, so just removed the method.

broofa (approved):

bluwy:
Thanks!

PR #318, opened 2025-03-19
fix: custom package sample code

mmkal:
The JS code for generating a URL for a custom package didn't work for me:

1. it uses `q=my_package` instead of `q=my_package@0.0.1` like the example above it - from experimenting it seems that this prevents it from using the supplied custom packageJson, at least in the case of existing libraries that exist on the npm registry (I'm trying to use this to get a graph for a package _before_ I release a new version)
2. it uses slightly non-intuitive `URLSearchParams` constructor -  I think the way I updated is more readable, with a plain JS object
3. the `q=my_package` part is hard-coded - until  https://github.com/npmgraph/npmgraph/issues/195 is resolve I think it makes sense to provide a JS snippet which does the name and version outputting for you
4. formatting of the code snippet is made slightly more readable (IMO) 

only 1 is a "fix"; 2, 3 and 4 are just niceties.


broofa:
Slow maintainers gonna slow, amiright? (sorry!)

Good catch on the documentation bug.  Thanks for taking the time to put up this PR.  Good call on the other improvements, too.

I've tweaked your code sample a bit.  (Fleshed it out to demonstrate supplying multiple packages, used `URL`, etc.)   I also took the liberty of adding documentation for a couple params that were missing from the docs.

PR #326, opened 2025-06-14
fix: detect and display dual-module support in colorizer

### LINKED ISSUE #325: specify if package is dual
currently, it looks like npmgraph decides whether a package is esm or not just by looking at the type field.

if the type field is `module`, even if the package provides cjs exports (as well as the esm ones) npmgraph will still show it's esm, potentially confusing cjs users that can't migrate to esm. to improve this, i propose adding a new "color" label to colorize for dual packages, something like this:

Image

broofa:
Fixes #325.

@SuperchupuDev: If you could do some testing for this, that would be helpful (see Vercel preview build link below).  The logic I've added for detecting ESM and CJS support is almost certainly imperfect.  It just walks the `package.json#exports` field looking for `import` or `require` fields, or paths with `.mjs` or `.cjs`.  I expect it's good enough to ship, but there's no real standard for how dual-module packages should be setup.  Thus, I'm sure there will be edge cases that aren't properly detected.

CleanShot 2025-06-13 at 23 15 57@2x

### DIFF
--- a/components/GraphPane/colorizers/ModuleTypeColorizer.tsx
+++ b/components/GraphPane/colorizers/ModuleTypeColorizer.tsx
@@ -17,16 +17,20 @@ export default {
         <LegendColor color={COLORIZE_MODULE_CJS}>CJS Only</LegendColor>
         <LegendColor color={COLORIZE_MODULE_DUAL}>CJS and ESM</LegendColor>
         <LegendColor color={COLORIZE_MODULE_ESM}>ESM Only</LegendColor>
-        <LegendColor color={COLORIZE_MODULE_TYPES}>TS Types</LegendColor>
+        <LegendColor color={COLORIZE_MODULE_TYPES}>
+          TypeScript types
+        </LegendColor>
       </>
     );
   },
 
   async colorForModule(module: Module) {
     const pkgType = detectPackageType(module.package as PackageJSON);
+
     if (pkgType.esm && pkgType.cjs) {
       return COLORIZE_MODULE_DUAL;
     }
+
     if (pkgType.esm) {
       return COLORIZE_MODULE_ESM;
     }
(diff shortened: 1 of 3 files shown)


SuperchupuDev:
hi, thank you for implementing this so fast! i can't find any dual package that doesn't use the `exports` field but i think the logic from https://github.com/antfu/node-modules-inspector/blob/a49ad3a91462249838d7b64b120e75e1357f41b0/packages/node-modules-tools/src/analyze-esm.ts#L8 is accurate if you want a reference implementation. might be worth adding a `faux` label too but i don't really mind :+1:

components/GraphPane/colorizers/ModuleTypeColorizer.tsx
```diff
@@ -12,14 +14,94 @@ export default {
...
+  if (exports && typeof exports === 'object') {
+    for (const [k, v] of Object.entries(exports)) {
+      if (k === 'import') {
+        pkgType.esm = true;
+      } else if (k === 'require') {
+        pkgType.cjs = true;
+      } else if (k === 'default') {
+        pkgType.esm = true;
+        pkgType.cjs = true;
+      }
```
SuperchupuDev, line 95: this doesn't sound right, if a package were to have only the default export it would only have one of them, but it would show up as dual

broofa: Fixed.  I read through the `node-modules.dev` and `npm-esm-vs-cjs` code a bit more closely and have tried to follow their lead on this. (I'm not directly pulling in the code from those projects, however, as it doesn't quite do what I'm after.)

broofa: FWIW: https://npmgraph-git-dualmodules-broofas-projects.vercel.app/?q=webdriver#color=moduleType&zoom=w seems to have a pretty good selection of module types... if you wanted to click through some of the `package.json` files for some of the modules to verify the detection logic works correctly.

(And, yes, I know there should be unit tests for this, but that's time and energy I don't have at the moment. :-) )

PR #380, opened 2026-05-03
feat: add header

### LINKED ISSUE #223: Keep search field visible
I think keeping the field visible at all times is useful:

- In the most common use case, it shows the name of the package being plotted. For huge projects this is helpful: https://npmgraph.js.org/?q=web-ext
- It could enable this: https://github.com/npmgraph/npmgraph/issues/222

The field used to be visible, as seen in an old PR: https://github.com/npmgraph/npmgraph/pull/91

fregante:
- closes https://github.com/npmgraph/npmgraph/issues/223
- replaces https://github.com/npmgraph/npmgraph/pull/378

Taking over the PR because it requires a lot of changes.


fregante:
@broofa leaving code aside since it's not done yet, thoughts on the current look?

I'm thinking of a couple of substantial behavior changes:

- tabs are always visible on the header
- a "closed" sidebar will appear as "no tabs selected"
- clicking on a tab will reopen the sidebar
- the "collapse sidebar" will likely become an "x" (or specific sidebar icon) in the corner and it will disappear when the closed

fregante:
It's looking pretty good! The only thing missing should be the mobile version, which will have to look somewhat different than before due to the relocation of the tabs.

Note that there's more work coming most likely in the Info pane (e.g. moving the GitHub link to the header, etc) but since this PR is already large, let's focus on the MVP for this substantial UI change.
 
Screenshot 1

Screenshot 3

index.scss
```diff
@@ -162,3 +147,35 @@ footer {
   color: var(--accent);
   text-decoration: underline;
 }
+
+.bright-hover {
```
fregante, line 151: I had to relocate some CSS initially due to some inconsistencies between CSS modules and non-CSS modules for CSS grid area names.

broofa (approved):
Approving, but this'll need to be updated to account for #379

fregante:
I think this is good to go. I'll let you re-review before we go live.

The mobile view matches the old one with the exception of the new header. I think I'll have to rethink it in a new PR anyway, it's always been a MVP.

broofa (approved):
This looks good.  Great work!

PR #381, opened 2026-05-04
chore: clean eslint config

fregante:
- https://github.com/antfu/eslint-config/issues/615


components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -372,7 +372,7 @@ function useGraphviz() {
   const [graphviz, setGraphviz] = useState<Graphviz | undefined>(undefined);
   const [loading, setLoading] = useState(true);
 
-  useMemo(() => {
+  useEffect(() => {
```
fregante, line 375: >  375:3  error  The return value of 'useMemo' must be assigned to a variable. Calling 'useMemo' without capturing its return value is likely a mistake — use 'useEffect' for side effects instead  react/use-memo

broofa (approved):

PR #389, opened 2026-05-05
chore: css modules throughout

broofa:


broofa:
Clicked around the UI a bunch... everything seems to be working.  Merging so as to avoid conflicts.

PR #390, opened 2026-05-06
chore: more css module work, fix printing and downloading

broofa:


fregante:
"Index" is not a module. Some style should remain global, like `.d-block` utility classes and html/body/a styles

PR #391, opened 2026-05-06
feat: add clear graph button

### LINKED ISSUE #257: Add button to clear graph (start from scratch)
If I want to start anew, my only way is to manually clear the URL.

It would be nice to handle `esc` and/or have a "clear" button somewhere.

What would be the action though? Should it reload `/` or just clean its internals? I occasionally have issues with the cache (#256) so I think the former would be best.

MVP:

```js
document.addEventListener('keypress', ev => {
  ev.keyCode === 'Esc' && location.href = '/'
})
```

fregante:
- Closes #257 

f

### DIFF
--- a/components/AppHeader.module.scss
+++ b/components/AppHeader.module.scss
@@ -24,6 +24,10 @@
     min-width: 0;
   }
 
+  a {
+    color: inherit;
+  }
+
   @include tight-screen {
     .tabs {
       display: none;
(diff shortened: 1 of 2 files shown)


fregante:
Pretty straightforward, it even auto-focuses the input field. Merging.

I also noticed that this was somewhat possible before by pressing in order: 
1. <kbd>/</kbd> to focus the input
2. <kbd>esc</kbd> to clear it (due to `type="search"`)
3. <kbd>enter</kbd> to submit an empty `q`

PR #400, opened 2026-05-06
feat: update to module-replacements v3 beta

ghostdevv:
We've been releasing module-replacements v3 beta's for a few months now and are going to cut the new major soon, just haven't had the chance :smile: It brings a bunch of new improvements to the data shape, especially for tooling that is aware of the environment as you can now see what replacements are available with those constraints. For this PR however I kept the change minimal, while also adding the minimum node version for native replacements as seems relevant. I also updated the module-replacements repo link as it's been moved to the e18e org.

Apologies if I made some JSX faux pas, I don't use it often :smile: 

image


broofa (approved):
This looks good, thanks!

PR #422, opened 2026-05-11
fix: don't auto-install commit hooks

fregante:
CI already exists to enforce conventions. `npx lefthook install` can still be run manually for those interested in pre-commit hooks rather than forcing them to use them. `git commit` is the same as `Save file` and saving files should not be blocked.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
   "scripts": {
     "build": "parcel build index.html robots.txt --no-cache",
     "fetch-licenses": "node scripts/fetch_licenses.js",
-    "prepare": "lefthook install",
     "release": "standard-version",
     "start": "parcel serve index.html robots.txt --no-cache",
     "lint": "eslint .",


broofa:
> `npx lefthook install` can still be run manually for those interested in pre-commit hooks rather than forcing them to use them.

'Kind of feel like I should put my foot down on this one.  These hooks are an early check to prevent committing code that breaks CI.  What reason is there for *not* having them installed?

fregante:
I cannot commit to `main`, so (pardon the french, it's just an expression) _what do you care_ about what I commit in secondary branches? CI is already what keeps repo clean, pre-commits duplicate the effort.

> What reason is there for _not_ having them installed?

1. Seriously dislike conventional commits and never use them
2. The only place where conventional commits matter is on the main branch, and it's fine by me, but again that's controlled at the PR level, `commitlint` can't affect that


but more specifically:

- I use commits to save the current state and then run auto-fixes or continue work separately. If your autofix breaks something or changes the behavior of what I wrote, I can't tell because, ahem, _my_ code was not committed. 

This is a serious drawback of pre-commit hooks, even if in practice this is only formatting the code, not changing it.

broofa (approved):
Fair enough!

PR #423, opened 2026-05-11
fix: drop URL workaround

fregante:
- includes https://github.com/npm/hosted-git-info/pull/334

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "d3-selection": "3.0.0",
     "d3-shape": "3.2.0",
     "filter-altered-clicks": "2.1.1",
-    "hosted-git-info": "^9.0.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "d3-selection": "3.0.0",
     "d3-shape": "3.2.0",
     "filter-altered-clicks": "2.1.1",
-    "hosted-git-info": "^9.0.2",
+    "hosted-git-info": "^10.1.0",
```
fregante, line 59: The major bump is:

> hosted-git-info now supports node ^22.22.2 || ^24.15.0 || >=26.0.0

which already matches our `devEngines`

fregante:
Tested, it works

PR #431, opened 2026-06-13
fix: refactor flash code

### LINKED ISSUE #430: More graceful failure for bad input via URL
## Repro

1. https://npmgraph.js.org/?q=undefined%40undefined#packages=%5B%7B%22dependencies%

The app fails to load entirely and logs an error in the console.

## Related

- https://github.com/npmgraph/npmgraph-cli/issues/5

broofa:
Bunch of lint and formatting fixes in here stemming from the switch to `xo`.

The issue in question was that the ad-hoc `flash()` logic was assuming the graph element was present, which it isn't early on in the page lifecycle.  Also, because the root element in React is now the document.body, it (React) was removing the flash elements any time it updated the DOM.  So I converted the flash code to a `Flash` component using AI (which, of course, makes for a more complicated implementation but... whatever.  I'm getting good at ignoring AI-induced complexity.)

fixes #430


components/GraphDiagram/GraphDiagram.tsx
```diff
@@ -398,6 +399,7 @@ function useGraphviz() {
   return [graphviz, loading] as const;
 }
 
+// eslint-disable-next-line complexity
```
fregante: If you're going to disable it for every instance it's best to just place this in the eslint config

lib/Module.ts
```diff
@@ -29,7 +29,7 @@ export default class Module {
     return module;
   }
 
-  // TODO: This should take either PackumentVersion or PackageJSON... but need to
+  // Note: This should take either PackumentVersion or PackageJSON... but need to
```
fregante: This looks like a TODO but renamed simply to avoid detection by the lint rule. Then let's just disable the rule globally

components/Tag.tsx
```diff
@@ -44,7 +44,7 @@ export function Tag({
     <div
       className={cn(
         styles.tag,
-        type === QueryType.Maintainer && styles.maintainer,
+        type === QueryType.Maintainer ? styles.maintainer : '',
```
fregante, line 47: ?

I don’t think this makes any difference?

components/Flash/Flash.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const graphWidth = graph instanceof HTMLElement ? graph.offsetWidth : 0;
+  const maxWidth = Math.max(
+    160,
+    graphWidth > 0 ? graphWidth - FLASH_GAP : window.innerWidth - FLASH_GAP * 2,
+  );
+
+  const appHeader = document
+    .getElementsByClassName(appHeaderStyles.root)
+    .item(0);
```
fregante: $('.' + appHeaderStyles.root)

components/Flash/Flash.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  let top = 0;
+  if (appHeader instanceof HTMLElement) {
```
fregante: In what scenario is it not found? Is this just a consequence of the nullish types? Using $() would avoid it

components/GraphDiagram/GraphDiagram.tsx
broofa: reverted and disabled.

lib/Module.ts
broofa: reverted and disabled

components/Tag.tsx
broofa: fixes TS complaint about type being `boolean | string` instead of just `string`

components/Flash/Flash.tsx
broofa: `$()` throws if it doesn't find an element

components/Flash/Flash.tsx
broofa: During the initial component render `graph` and `appHeader` may not be in the DOM yet. 🤷

